PR #20, opened 2023-02-01
feat: Properly allow CSS overrides for Supaglue components

asdfryan:
This PR allows to tailwind (and other css) overrides to Supaglue components without forcing it via the `!important` tag.

This includes:
* Migration from using css modules to emotion styles, whose `CacheProvider` allows for proper injection of default stylings that can still be overridden
* Removes `type` attribute from buttons so default stylings do not get reset by tailwind globals
* Light refactoring of providers / some components

Future:
* Emotion styles should be refactored to maximize re-usability (`ThemeProvider` maybe?) and performance


packages/nextjs/src/lib/styles.ts
```diff
@@ -0,0 +1,14 @@
...
+  SLATE_8: 'hsl(205 10.7% 78.0%)',
+  SLATE_9: 'hsl(226 58.6% 51.3%)',
+  SLATE_12: 'hsl(206 24.0% 9.0%)',
+  INDIGO_5: 'hsl(224 87.1% 92.0%)',
+  INDIGO_6: 'hsl(224 81.9% 87.8%)',
+  INDIGO_9: 'hsl(226 70.0% 55.5%)',
+  INDIGO_10: 'hsl(226 58.6% 51.3%)',
+  INDIGO_12: 'hsl(226 62.0% 17.0%)',
+  GRAY_12: 'hsl(0, 0%, 93.0%)',
+} as const;
```
khennes, line 14: Radix Colors should work out of the box with emotion, so we don't have to hardcode the scales anymore: https://www.radix-ui.com/docs/colors/getting-started/usage#emotion

khennes (approved):
nice! ⚡

packages/nextjs/src/components/TriggerSyncButton/TriggerSyncButton.tsx
```diff
@@ -53,19 +54,18 @@ export const TriggerSyncButtonInternal = ({
   };
 
   return integrationConnected ? (
-    <button
-      className={classNames('sg-buttonLabel', appearance?.elements?.buttonLabel, styles.button)}
+    <Button
+      className={classNames('sg-triggerSyncButton', appearance?.elements?.button)}
```
tomkit, line 58: Let's chat about our namespaced nomenclature for classes

packages/nextjs/src/providers/supaglueProviderInternal.tsx
```diff
@@ -0,0 +1,31 @@
+import createCache from '@emotion/cache';
```
tomkit, line 1: I think react file convention here is Pascal case?

packages/nextjs/src/lib/styles.ts
asdfryan: TIL -- thanks!

tomkit (commented):
Thx Ryan

packages/nextjs/src/providers/supaglueProviderInternal.tsx
asdfryan: Even if it's a function?

packages/nextjs/src/components/TriggerSyncButton/TriggerSyncButton.tsx
asdfryan: Chatted offline -- removing everything except primitives (i.e. `sg-label`, `sg-button`, etc)

asdfryan: just kidding -- we decided to keep this (and maybe add them to every component we export)

PR #30, opened 2023-02-03
chore(todos): link some todos with tickets. remove unused appearance …

tomkit:
…interface. pascal case provider

### DIFF
--- a/apps/api/routes/index.ts
+++ b/apps/api/routes/index.ts
@@ -7,6 +7,7 @@ import integration from './integration';
 import oauth from './oauth';
 import sync from './sync';
 
+// TODO: ENG-105 version and guard routes
 export default function initRoutes(app: Express): void {
   app.use('/developer_config', developerConfig);
   app.use('/integrations', integration);
--- a/packages/nextjs/src/types/index.ts
+++ b/packages/nextjs/src/types/index.ts
@@ -1,29 +1,3 @@
 export type SupaglueAppearance = {
-  theme?: 'light';
-  layout?: {
-    logoPlacement: 'inside' | 'outside' | 'none';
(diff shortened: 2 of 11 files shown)


apps/api/routes/index.ts
```diff
@@ -7,6 +7,7 @@ import integration from './integration';
 import oauth from './oauth';
 import sync from './sync';
 
+// TODO: ENG-105 version and guard routes
```
lucasmarshall, line 10: Probably should reference the GitHub issue instead

packages/nextjs/src/types/index.ts
```diff
@@ -1,29 +1,26 @@
...
-    colorInputBackground?: string;
-    fontFamily?: string;
-    fontFamilyButtons?: string;
-    fontSize?: string;
-    fontSmoothing?: string;
-    fontWeight?: string;
-    borderRadius?: string;
-    spacingUnit?: string;
-  };
+  // TODO: ENG-109
```
lucasmarshall, line 2: Same as above.

lucasmarshall (changes requested):

apps/api/routes/index.ts
tomkit: The synced GH ticket has the Linear ENG-xx as well

packages/nextjs/src/types/index.ts
```diff
@@ -1,29 +1,26 @@
...
+  // theme?: 'light';
```
asdfryan, line 3: I would actually just remove these for now since we're still not sure what the interface will look like and don't want to preview anything we're not sure of.

asdfryan (approved):

PR #31, opened 2023-02-03
chore(docs): short section on integrations sdk

tomkit:


docs/versioned_docs/version-0.1.0/integrations/salesforce.md
```diff
@@ -190,3 +192,13 @@ sdk.config({
   defaultFieldMapping: contactMapping,
 });
 ```
+
+## Integrations SDK
+
+Internally, Supaglue uses an evolving Integrations SDK that wraps Salesforce and Supaglue Destinations so we can imperatively write business logic for Syncs, which then later get invoked by developers using the Config SDK above. In the future, we can expose the Integrations SDK to developers for them to define their own tailored business logic for Syncs. The rough shape of calls using it looks like:
```
lucasmarshall, line 198: What is a "Supaglue Destination"?

tomkit: can simply be `Destinations` under /concepts/destinations

lucasmarshall: Not sure it's clear here. Link?

albertyfwu (approved):

PR #38, opened 2023-02-03
fix: make setup easier

lucasmarshall:
Also fixes a few things that were incorrect about how we handle the api url in the nextjs components and sample-app.


.env.sample
```diff
@@ -1,5 +1,3 @@
-NEXT_PUBLIC_SUPAGLUE_HOST=http://localhost:8080
```
lucasmarshall, line 1: This wasn't read by anything.

README.md
```diff
@@ -48,7 +48,7 @@ We are currently in Public Alpha. Watch "releases" of this repo to be notified o
 ```shell
 git clone git@github.com:supaglue-labs/supaglue.git
 cd supaglue
-cp .env.sample .env
```
lucasmarshall, line 51: We have defaults in the docker compose that makes this not nessessary

docs/docusaurus.config.js
```diff
@@ -33,12 +33,8 @@ const config = {
       /** @type {import('@docusaurus/preset-classic').Options} */
       ({
         docs: {
-          onlyIncludeVersions: ['0.1.0'],
```
lucasmarshall, line 36: Show "Next" version of docs.

packages/nextjs/src/providers/supaglueProvider.tsx
```diff
@@ -7,7 +7,7 @@ const SupaglueContext = createContext({
 
 type SupaglueProviderProps = {
   children: ReactNode;
-  apiUrl: string;
```
lucasmarshall, line 10: Should be optional since we can fallback to the env var.

khennes (dismissed):

PR #39, opened 2023-02-03
fix: add SWRConfig to SupaglueProvider for useSalesforceIntegration hook

albertyfwu:
Without this, the `useSalesforceIntegration` wasn't making a network call and therefore always redirecting away from `/integrations/salesforce` to `/integrations` on load.


khennes:
> the useSalesforceIntegration wasn't making a network call 

seems like the issue could be that there's an existing cache entry for this endpoint?

albertyfwu:
> > the useSalesforceIntegration wasn't making a network call
> 
> seems like the issue could be that there's an existing cache entry for this endpoint?

The hook wasn't returning any data no matter what (even if there potentially was a cache entry). `isLoading = false` and `data = undefined` always

khennes (approved):

PR #42, opened 2023-02-03
feat: show actual cron schedule in Switch component

khennes:
Found this helpful while working on incremental sync, so decided to check it in.

- Adds a new boolean prop, `includeSyncDescription`, to the `Switch` component.
- Also adds a new targetable `element` to the `appearance` prop (verified that overriding the `switchDescription` styles works 👍)
- Moves responsibility for displaying the sync description from the client to the component lib
- Adds a new dependency, [cronstrue](https://github.com/bradymholt/cRonstrue), to the `nextjs` package. Normally I'm leery of adding dependencies like this, but in this case, it's quick and easy to use, its bundle size is tiny, and it has no dependencies itself. Also, it has a cute name 💁‍♀️ 

Screen Shot 2023-02-03 at 2 09 35 PM

Review [with whitespace removed](https://github.com/supaglue-labs/supaglue/pull/42/files?diff=unified&w=1) for an easier time.


lucasmarshall (dismissed):

asdfryan (approved):
Nice!

Issue #17, opened 2023-02-01
[ENG-109] Epic:`theme` customizations for React components


tomkit:
<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-109](https://linear.app/supaglue/issue/ENG-109/epictheme-and-variables-customizations-for-react-components)</sub>

asdfryan:
This is now available in `main`, to be included in the `0.2.0` release.

We have provided 2 themes (`defaultTheme` and `darkTheme`) that can be passed into the `SupaglueProvider`. 

```
// TUTORIAL: Uncomment this
import { darkTheme } from '@supaglue/nextjs';
// ...
 <SupaglueProvider
   customerId={customerId}
   // TUTORIAL: Uncomment this
   theme={darkTheme}
 >
```

![image](https://user-images.githubusercontent.com/1498449/217958765-3de827ed-c680-4dd9-8081-d6c965d467fb.png)

Furthermore, the `theme` class can be further overridden to change background colors, primary/danger color schemes, or even element-specific overrides.

PR #43, opened 2023-02-03
feat: implement overridable light and dark themes with variant support

asdfryan:
* theme is an optional prop into `SupaglueProvider` (by default, it will use the light theme)
* instead of passing in interpolated css prop, pass in functions taking `theme` as prop into each component
* Support for variants (right now `primary` and `danger` variants for Buttons)
* Fixes https://github.com/supaglue-labs/supaglue/issues/17

Screenshot 2023-02-03 at 2 20 49 PM
Screenshot 2023-02-03 at 2 20 35 PM
Screenshot 2023-02-03 at 2 32 02 PM


tomkit (approved):

asdfryan:
> I spent a few hours trying to figure out the types issue here and just can't. I think we'll have to do what clerk does and cast to any.

Thanks -- I'll take a look at what I can do to at least make the developer experience more friendly.

packages/nextjs/src/style/types/variant.ts
```diff
@@ -0,0 +1,34 @@
...
+  [key in keyof V]?: UnwrapBooleanVariant<keyof V[key]>;
+};
+
+export type DefaultVariants = Record<string, string>;
+
+export type CreateVariantsConfig<V> = {
+  base?: StyleRule;
+  variants: V;
+  defaultVariants?: DefaultVariants;
+  elementOverrides?: Record<string, any>;
```
tomkit, line 22: quick comment

packages/nextjs/src/style/internal/applyTheme.ts
```diff
@@ -0,0 +1,6 @@
+import { SgTheme } from '../types/theme';
```
tomkit, line 1: quick comment

tomkit (dismissed):

tomkit (approved):

PR #46, opened 2023-02-03
fix(sample-app): fix api url

lucasmarshall:
Also update turbo


packages/nextjs/src/providers/SupaglueProvider.tsx
```diff
@@ -26,6 +26,7 @@ export const SupaglueProvider: FC<SupaglueProviderProps> = ({ children, customer
       <SWRConfig
         value={{
           fetcher: async (config: RequestType) => {
+            const { apiUrl } = context;
```
lucasmarshall, line 29: This is the fix

khennes (approved):

PR #55, opened 2023-02-05
feat: add SalesforceDisconnectButton component

khennes:
Adds a button to disconnect a Salesforce account and delete the end user's integration.

Screenshot 2023-02-06 at 12 04 33 PM

TO DO (later):
- Right now, clicking "Disconnect" both logs the user out and deletes the SFDC integration. In the future, we might allow users to keep the integration but disconnect in order to switch the Salesforce account they use to authenticate
- Currently the `DELETE /integrations/:id` endpoint takes a _long_ time and gives no feedback to the user - we should add a button loading state and consider offloading the Temporal schedule deletions to an async task
- ~See if there's an alternative to appending `prompt=login` to the SFDC authz url.~ (Currently, without it, a user gets automatically logged in after clicking "Disconnect" and then "Connect" again. @lucasmarshall do you know why this might be? Something to do with Salesforce session policies maybe?)

### DIFF
--- a/apps/api/integrations/services/integration_service.ts
+++ b/apps/api/integrations/services/integration_service.ts
@@ -22,7 +22,10 @@ export class IntegrationService {
     this.#syncService = syncService;
   }
 
-  public async getById(id: string): Promise<SafeIntegration> {
+  public async getById(id: string, unsafe?: false): Promise<SafeIntegration>;
+  public async getById(id: string, unsafe?: true): Promise<UnsafeIntegration>;
+  public async getById(id: string, unsafe?: undefined): Promise<SafeIntegration>;
+  public async getById(id: string, unsafe = false): Promise<SafeIntegration | UnsafeIntegration> {
     const integration = await this.#prisma.integration.findUnique({
       where: {
         id,
@@ -31,6 +34,10 @@ export class IntegrationService {
     if (!integration) {
       throw new Error(`Can't find integration with id: ${id}`);
     }
+    if (unsafe) {
+      return fromModelToUnsafeIntegration(integration);
+    }
+
     return fromModelToSafeIntegration(integration);
   }
 
@@ -83,5 +90,15 @@ export class IntegrationService {
     }
 
     return fromModelToSafeIntegration(integration);
+  }
+
+  public async delete(integrationId: string): Promise<void> {
+    return await this.#prisma.$transaction(async (tx) => {
+      const integration = await tx.integration.findUniqueOrThrow({ where: { id: integrationId } });
+      await Promise.all([
+        this.#syncService.deleteSyncsForCustomer(tx, integration.customerId),
+        tx.integration.delete({ where: { id: integration.id } }),
+      ]);
+    });
   }
 }
--- a/apps/api/routes/integration.ts
+++ b/apps/api/routes/integration.ts
@@ -1,9 +1,10 @@
 import { Request, Response, Router } from 'express';
+import jsforce from 'jsforce';
 import { getDependencyContainer } from '../dependency_container';
 import { IntegrationCreateParams, SafeIntegration } from '../integrations/entities';
 import { errorMiddleware as posthogErrorMiddleware, middleware as posthogMiddleware } from '../lib/posthog';
 
-const { integrationService } = getDependencyContainer();
+const { integrationService, developerConfigService } = getDependencyContainer();
 
 const router: Router = Router({ mergeParams: true });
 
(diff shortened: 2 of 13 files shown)


apps/api/syncs/services/sync_service.ts
```diff
@@ -191,6 +191,27 @@ export class SyncService {
...
   }
 
+  public async deleteSyncsForCustomer(customerId: string): Promise<void> {
+    const syncs = await this.#prisma.sync.findMany({
+      where: { customerId },
+      select: { id: true, enabled: true },
+    });
+
+    // Pause all enabled syncs right away
+    await Promise.all(syncs.filter((sync) => sync.enabled).map((enabledSync) => this.pauseSync(enabledSync.id)));
```
asdfryan, line 201: What's the point of this?

packages/nextjs/src/components/Button/styles.ts
```diff
@@ -1,27 +1,44 @@
...
+    cursor: 'auto',
+  },
+});
+
+const destructiveButton = css(button, {
+  backgroundColor: red.red9,
+  ':hover': {
+    backgroundColor: red.red10,
+  },
+  ':disabled': {
```
asdfryan, line 33: Looking ahead to theming -- I had thought we would use the same `disabled` styling for all variants (primary, danger etc.). Do you have a preference?

apps/api/integrations/services/integration_service.ts
```diff
@@ -84,4 +91,11 @@ export class IntegrationService {
 
     return fromModelToSafeIntegration(integration);
   }
+
+  public async delete(integrationId: string): Promise<void> {
+    // TODO: Wrap in transaction
+    const integration = await this.#prisma.integration.findUniqueOrThrow({ where: { id: integrationId } });
+    await this.#syncService.deleteSyncsForCustomer(integration.customerId);
```
asdfryan, line 98: Should we parallelize these 2?

apps/api/routes/integration.ts
```diff
@@ -42,4 +42,28 @@ router.post(
...
+router.delete(
+  '/:integrationId',
+  posthogMiddleware('Delete Integration'),
+  async (req: Request<{ integrationId: string }>, res: Response<Record<string, never>>) => {
+    const integration = await integrationService.getById(req.params.integrationId, /* unsafe */ true);
+    const { refreshToken } = integration.credentials;
+    const developerConfig = await developerConfigService.getDeveloperConfig();
+
+    // Log out of Salesforce first
+    const oauth2 = new jsforce.OAuth2({
```
asdfryan, line 54: Non-blocking: Just for my own curiosity, Is this idempotent? What happens if we call this twice?

apps/api/syncs/services/sync_service.ts
khennes: Just a precaution to ensure that no scheduled syncs start running while we're deleting the integration. It's also a safeguard in case the schedule deletions (line 208) fail for some reason.

packages/nextjs/src/components/Button/styles.ts
khennes: Good call, I'll remove this

apps/api/integrations/services/integration_service.ts
khennes: Sure, will update

apps/api/routes/integration.ts
khennes: Do you mean line 59? If we call the logout endpoint twice, the second request *should* fail with a 400 (because the token is invalid). I didn't test logging out twice, but I did test revoking the token twice.

asdfryan: I guess what I meant is perhaps if the user is already logged out, we should just skip this step instead of trying to log out (and then throwing an error).

But this is non-blocking.

asdfryan (approved):

lucasmarshall:
> Adds a button to disconnect a Salesforce account and delete the end user's integration.
> 
> Screenshot 2023-02-06 at 12 04 33 PM
> 
> TO DO (later):
> 
> * Right now, clicking "Disconnect" both logs the user out and deletes the SFDC integration. In the future, we might allow users to keep the integration but disconnect in order to switch the Salesforce account they use to authenticate
> * Currently the `DELETE /integrations/:id` endpoint takes a _long_ time and gives no feedback to the user - we should add a button loading state and consider offloading the Temporal schedule deletions to an async task
> * See if there's an alternative to appending `prompt=login` to the SFDC authz url. (Currently, without it, a user gets automatically logged in after clicking "Disconnect" and then "Connect" again. @lucasmarshall do you know why this might be? Something to do with Salesforce session policies maybe?)

This is pretty common in a lot of OAuth implementations. Basically, if you haven't actually gone and revoked the OAuth session on Salesforce's side, it resumes the "session" it had before.

apps/api/routes/integration.ts
khennes: Ah yeah, good point, added a todo for this.

lucasmarshall (approved):

khennes:
Fixes ENG-137

PR #56, opened 2023-02-06
feat: basic incremental sync

khennes:


khennes:
(per @tomkit, pausing this until we ship the new data model)

PR #57, opened 2023-02-06
docs(fix): remove superfluous key prop for docs

tomkit:


khennes (commented):
Without the key passed to `FieldMapping` users will see this problem again: https://linear.app/supaglue/issue/ENG-77/switching-tabs-under-salesforce-integration-causes-field-map-values-to

asdfryan:
Yeah, we should revert this IMO.

khennes:
Alternatively this should work: #65

PR #66, opened 2023-02-06
docs: reference code blocks by name, fix ordering, make bullets

asdfryan:


docs/versioned_docs/version-0.1.x/tutorial.md
```diff
@@ -92,11 +92,11 @@ For this tutorial, we've included a sample [Developer Config](/concepts#develope
...
    ```
 
-   The `syncConfig` function creates a Sync Config that would allow customers to pull all Account records from their Salesforce instance into the sample app's Postgres database every 15 minutes.
+   - The `accountsSchema` object exposes the specified Salesforce Account fields to your customers and lets them map them to fields in the sample app's Postgres database.
 
-   The schema object exposes the specified Salesforce Account fields to your customers and lets them map them to fields in the sample app's Postgres database.
+   - The `defaultFieldMapping` object specifies the default field mapping values for the Salesforce Account object that would be used in the sample app absent any customer-provided overrides.
 
-   The defaultFieldMapping object specifies the default field mapping values for the Salesforce Account object that would be used in the sample app absent any customer-provided overrides.
+   - The `syncConfig` function creates a Sync Config that would allow customers to pull all Account records from their Salesforce instance into the sample app's Postgres database every 15 minutes.
```
khennes, line 99: should we mention that "every 15 minutes" is customizable?

khennes (approved):

khennes (approved):

PR #68, opened 2023-02-07
feat(api): endpoint to get sync run logs

lucasmarshall:
Screenshot 2023-02-07 at 12 53 11 PM

### DIFF
--- a/apps/api/routes/sync.ts
+++ b/apps/api/routes/sync.ts
@@ -1,7 +1,7 @@
 import { Request, Response, Router } from 'express';
 import { getDependencyContainer } from '../dependency_container';
 import { errorMiddleware as posthogErrorMiddleware, middleware as posthogMiddleware } from '../lib/posthog';
-import { Sync, SyncCreateParams } from '../syncs/entities';
+import { Sync, SyncCreateParams, SyncRun, SyncRunStatus } from '../syncs/entities';
 
 const { syncService, developerConfigService } = getDependencyContainer();
 
@@ -77,6 +77,27 @@ router.post(
     return res.status(200).send();
   },
   posthogErrorMiddleware('Manually Trigger Sync')
+);
+
+router.get(
+  '/run_logs',
+  posthogMiddleware('Get Sync Run Logs'),
+  async (
+    req: Request<
+      never,
+      any,
+      never,
+      { syncConfigName?: string; customerId?: string; status?: SyncRunStatus; page?: number; count?: number }
+    >,
+    res: Response<{ logs: (SyncRun & { sync: Partial<Sync> })[] }>
+  ) => {
+    const { syncConfigName, customerId, status, page = 0, count = 10 } = req.query;
+
+    const logs = await syncService.getSyncRunLogs({ syncConfigName, customerId, status, page, count });
+
+    return res.status(200).send({ logs });
+  },
+  posthogErrorMiddleware('Get Sync Run Logs')
 );
 
 export default router;
--- a/apps/api/syncs/services/sync_service.ts
+++ b/apps/api/syncs/services/sync_service.ts
@@ -3,7 +3,15 @@ import { Client, ScheduleAlreadyRunning, ScheduleNotFoundError, ScheduleOverlapP
 import { DeveloperConfig, SyncConfig } from '../../developer_config/entities';
 import { TEMPORAL_SYNC_TASKS_TASK_QUEUE } from '../../temporal';
 import { getRunSyncWorkflowId, runSync } from '../../temporal/workflows';
-import { fromModelToSyncRun, Sync, SyncCreateParams, SyncRun, SyncUpdateParams } from '../entities';
+import {
+  fromModelToSyncRun,
+  fromModelToSyncRunWithSyncData,
+  Sync,
+  SyncCreateParams,
+  SyncRun,
+  SyncRunStatus,
+  SyncUpdateParams,
+} from '../entities';
 import { getSyncId } from '../util';
 
 export class SyncService {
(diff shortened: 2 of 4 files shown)


apps/api/syncs/services/sync_service.ts
```diff
@@ -222,10 +230,33 @@ export class SyncService {
   public async finishSyncRun(syncRunId: string, errorMessage?: string): Promise<SyncRun> {
     const model = await this.#prisma.syncRun.update({
       where: { id: syncRunId },
-      data: { finishTimestamp: new Date(), status: errorMessage ? 'error' : 'success' },
+      data: { finishTimestamp: new Date(), status: errorMessage ? 'error' : 'success', errorMessage },
```
lucasmarshall, line 233: We weren't actually recording the error message in the db

apps/api/temporal/activities/do_sync.ts
```diff
@@ -26,6 +27,8 @@ export async function doSync({ syncId, syncRunId }: DoSyncArgs): Promise<void> {
 async function doSyncImpl(sg: Supaglue, sync: Sync, syncConfig: SyncConfig, syncRunId: string): Promise<void> {
   const fieldMapping = getMapping(sync, syncConfig);
 
+  throw ApplicationFailure.nonRetryable('Sample error');
+
```
lucasmarshall, line 31: This was for testing. Will remove

apps/api/temporal/workflows/run_sync.ts
```diff
@@ -30,9 +30,10 @@ export async function runSync({ syncId }: RunSyncArgs): Promise<void> {
       err.cause instanceof ApplicationFailure &&
       err.cause.nonRetryable
     ) {
-      await finishSyncRun({ syncRunId, errorMessage: err.message });
-      await pauseSync({ syncId, note: err.message });
-      return;
```
lucasmarshall, line 35: We were returning here, so errors weren't getting logged in temporal correctly. We should still throw to help with debugging, metrics, etc.

apps/api/temporal/workflows/run_sync.ts
```diff
@@ -30,9 +30,10 @@ export async function runSync({ syncId }: RunSyncArgs): Promise<void> {
...
+      const errorMessage =
+        err.cause.failure?.message ?? err.failure?.cause?.message ?? err.failure?.message ?? err.message;
```
lucasmarshall, line 34: There are a lot of places where temporal errors can have messages, so run down them all.

apps/api/routes/sync.ts
```diff
@@ -79,4 +79,25 @@ router.post(
   posthogErrorMiddleware('Manually Trigger Sync')
 );
 
+router.get(
+  '/logs',
+  posthogMiddleware('Get Sync Logs'),
```
tomkit, line 84: If above is accurate, then `Sync` --> `Sync Run`

apps/api/routes/sync.ts
```diff
@@ -79,4 +79,25 @@ router.post(
...
+    >,
+    res: Response
+  ) => {
+    const { syncConfigName, customerId, status, page = 0, count = 10 } = req.query;
+
+    const logs = await syncService.getSyncRuns({ syncConfigName, customerId, status, page, pageCount: count });
+
+    return res.status(200).send({ logs });
+  },
+  posthogErrorMiddleware('Get Sync Logs')
```
tomkit, line 100: same as above

apps/api/routes/sync.ts
```diff
@@ -79,4 +79,25 @@ router.post(
   posthogErrorMiddleware('Manually Trigger Sync')
 );
 
+router.get(
+  '/logs',
```
tomkit, line 83: We're getting sync run logs, not sync logs, correct?

apps/api/temporal/workflows/run_sync.ts
```diff
@@ -30,9 +30,14 @@ export async function runSync({ syncId }: RunSyncArgs): Promise<void> {
...
     ) {
-      await finishSyncRun({ syncRunId, errorMessage: err.message });
-      await pauseSync({ syncId, note: err.message });
-      return;
+      const errorMessage =
+        err.cause.failure?.message ??
+        err.failure?.cause?.message ??
+        err.failure?.message ??
+        err.message ??
+        'Unknown error';
```
lucasmarshall, line 38: There are a lot of places where temporal errors have messages, so run down them all, but we also need a default if we can't find a message.

apps/api/routes/sync.ts
lucasmarshall: Good catch. Will change.

apps/api/temporal/workflows/run_sync.ts
```diff
@@ -30,9 +31,15 @@ export async function runSync({ syncId }: RunSyncArgs): Promise<void> {
...
+      await finishSyncRun({ syncRunId, errorMessage });
+      await pauseSync({ syncId, note: errorMessage });
+      logger.error(`Sync run ${syncRunId} failed: ${errorMessage}`, { syncId });
```
lucasmarshall, line 42: Also log to the container logs

apps/api/routes/sync.ts
```diff
@@ -79,4 +79,25 @@ router.post(
   posthogErrorMiddleware('Manually Trigger Sync')
 );
 
+router.get(
+  '/run_logs',
```
albertyfwu, line 83: Should we nest the sync runs under `/syncs` instead? e.g. `/syncs/:syncId/runs/:runId`?

lucasmarshall: We don't want it under `:syncId` since we want to be able to search by customer id, sync config name, status, or any combination of those. So we don't know the syncId or runId.

albertyfwu: Should we not even nest it under `/syncs` then and move it to top-level?

I don't feel too strongly about this, especially since it's for internal consumption anyway.

albertyfwu (approved):

PR #72, opened 2023-02-07
feat!: inbound sync types

albertyfwu:
This change involves adding a non-nullable column `type` to `syncs` table. If you already have `syncs`, you may want to truncate your `syncs` table before applying this migration. Alternatively, reset your environment with `docker compose down -v`.

### DIFF
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -34,6 +34,7 @@ model DeveloperConfig {
 model Sync {
   id                                String    @id @default(uuid())
   customerId                        String    @map("customer_id")
+  type                              String
   enabled                           Boolean
   syncConfigName                    String    @map("sync_config_name")
   fieldMapping                      Json?     @map("field_mapping")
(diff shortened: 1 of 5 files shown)


apps/api/prisma/schema.prisma
```diff
@@ -34,6 +34,7 @@ model DeveloperConfig {
 model Sync {
   id                                String    @id @default(uuid())
   customerId                        String    @map("customer_id")
+  type                              String
```
lucasmarshall, line 37: Probably should make this nullable? Otherwise the migration will fail for current users.

lucasmarshall: Also, we should generate the migration in this PR probably.

albertyfwu: Can we just make breaking changes to limit complexity of our data model? Nobody is using this in production right now.

Just pushed the migration. Forgot to push it :)

lucasmarshall: We probably should document how users can reset their DB if we are going to have breaking changes.

albertyfwu: I changed the conventional commit prefix to breaking and added documentation in this PR. Would that be sufficient given our current stage?

lucasmarshall (approved):

PR #74, opened 2023-02-07
feat: introduce outbound sync types in SDK

albertyfwu:
introducing outbound sync types in SDK. not usable yet


apps/api/developer_config/entities/developer_config.ts
```diff
@@ -36,14 +33,19 @@ export type PostgresDestination = BaseDestination & {
...
     };
     table: string;
-    upsertKey: string;
     customerIdColumn: string;
   };
 };
 
+export type PostgresDestination = PostgresInternalIntegration & {
+  config: {
+    upsertKey: string;
```
albertyfwu, line 42: only relevant when postgres is a destination

packages/sdk/src/base.ts
```diff
@@ -1,66 +1,54 @@
-import { DefaultFieldMapping } from './defaultFieldMapping';
+import { FieldMapping } from './defaultFieldMapping';
```
albertyfwu, line 1: We had an offline discussion about the types/classes in `packages/sdk` and understood that the current implementation of classes is for future-proofing.

After that, while I was copying over types from `api`, I found it to be very cumbersome and time-consuming to write out classes, `toJSON()` methods, etc. to mirror each type (esp. with union types, etc.).

I think keeping the functions around (e.g. `sdk.salesforce.inboundSyncConfig`) but making them no-ops that just return the params passed to them is a good middle-ground for now to retain extensibility in the future but optimize for speed now when we quickly iterate on types.

tomkit (approved):

PR #79, opened 2023-02-08
fix(sample-app): title on integrations page

lucasmarshall:


apps/sample-app/pages/integrations/[type].tsx
```diff
@@ -9,21 +9,24 @@ import { useActiveTab } from '../../hooks';
...
 
 export default function Integration() {
+  const router = useRouter();
+  const type = router.query.type as string;
+  const typeCaps = type ? type.charAt(0).toUpperCase() + type.slice(1) : '';
   return (
     <>
       <Head>
-        <title>Supaglue</title>
+        <title>ApollaIO - Integration - {typeCaps}</title>
```
Comment (tomkit) [inline apps/sample-app/pages/integrations/[type].tsx:18]: nit: can just do `Integration` or `Integration - Salesforce` for simplicity right now

tomkit (approved):
Left a nit comment

PR #84, opened 2023-02-09
docs(intro): clarify use case

tomkit:
- unnecessary admonition
- add outbound use case to next docs

### DIFF
--- a/docs/versioned_docs/version-0.1.x/intro.md
+++ b/docs/versioned_docs/version-0.1.x/intro.md
@@ -44,7 +44,7 @@ sources={{
   }}
 />
 
-- **Config SDK** to define author [syncs](/concepts#developer-config) declaratively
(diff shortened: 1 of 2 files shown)


docs/versioned_docs/version-0.1.x/intro.md
```diff
@@ -53,16 +53,10 @@ sources={{
 
 ## Use cases
 
-:::info
+Supaglue can be used by B2B SaaS companies to provide customer-facing Salesforce integrations as part of their products.
```
albertyfwu, line 56: shouldn't we only update the `next` docs? It is still true that outbound syncs are not supported in 0.1.x

george-xing (approved):

docs/versioned_docs/version-0.1.x/intro.md
tomkit: yes, only changed the wording here to be B2B SaaS in general

PR #85, opened 2023-02-09
chore: move toast to TriggerSyncButton component

asdfryan:
Right now the toast logic is written inside the sample app.

This is step 1 of moving the entire config setting into nextjs package.

We should consider making the toast a primitive, but it's tricky because it depends on its parent container. We should probably have a `NotificationProvider` that manages all of these, but punting for now.


tomkit:
> We should consider making the toast a primitive, but it's tricky because it depends on its parent container.
Agreed it should be just part of the SupaglueSyncConfig component for right now and punt on it being a primitive. I think a SyncLog component would probably be most impactful next

tomkit (approved):

PR #86, opened 2023-02-09
feat: Allow developer to pick 1 sObject or allow customer to choose from a list; also refactor

albertyfwu:
In adding the option on `SyncConfig` to allow the developer to pick 1 sObject for syncing or allow customer to choose from list (e.g. Contact vs Lead), I did a refactor to clean up the types a bit and put things where they belong.

For example, `source` and `destination` were used for inbound syncs, but they have opposite meanings for outbound syncs. Restructuring things a bit makes things less confusing.

**After this PR merges, you'll need to `supaglue apply <config>` again as the types have changed.**


apps/sample-app/supaglue-config/inbound/contact.ts
```diff
@@ -27,11 +27,13 @@ const contactSchema = sdk.schema({
   ],
 });
 
-const contactSyncConfig = sdk.salesforce.inboundSyncConfig({
+const contactSyncConfig = sdk.syncConfigs.inbound({
   name: 'Contacts',
-  salesforceObject: 'Contact',
+  source: sdk.customer.sources.salesforce({
+    objectConfig: sdk.customer.specifiedSalesforceObjectConfig('Contact'),
```
tomkit, line 33: it feels like external developer friction to have to specify it in this way

apps/sample-app/supaglue-config/inbound/contact.ts
```diff
@@ -27,11 +27,13 @@ const contactSchema = sdk.schema({
   ],
 });
 
-const contactSyncConfig = sdk.salesforce.inboundSyncConfig({
+const contactSyncConfig = sdk.syncConfigs.inbound({
```
tomkit, line 30: rather than the directionality being namespaced in the sdk, should it be specified as a member field of SyncConfig?

tomkit (commented):
what type differences are there between internal/external? 

the matrix of internal/external x inbound/outbound x sources/destinations is become a bit unwieldy as i look at the devxp

tomkit (approved):
We chatted offline: the interface sounds good for now, there are two questions around SDK scoping/namespacing we'll revisit

PR #91, opened 2023-02-09
fix: use field labels, rather than names, in FieldMapping dropdowns

khennes:
When using field names, custom fields are displayed with a `__c` suffix, e.g. `FavoriteFruit__c`.

### DIFF
--- a/packages/nextjs/src/components/FieldMapping/FieldMapping.tsx
+++ b/packages/nextjs/src/components/FieldMapping/FieldMapping.tsx
@@ -16,6 +16,11 @@ type Field = {
   value: string;
 };
 
+type SalesforceField = {
+  name: string;
+  label: string;
+};
+
 type FieldMapping = Record<string, string>;
 
 const EmptyContent = ({ className, children }: { className?: string; children: ReactNode }) => {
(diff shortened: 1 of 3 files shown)


albertyfwu (commented):
Is it possible in Salesforce to create a custom field with `Email` label and `Email__c` name even though an existing property `Email` already exists?

khennes:
> Is it possible in Salesforce to create a custom field with `Email` label and `Email__c` name even though an existing property `Email` already exists?
> 
> Does salesforce enforce uniqueness on field labels in addition to field names?

Apparently not:

Screenshot 2023-02-09 at 10 28 53 AM

albertyfwu:
> > Is it possible in Salesforce to create a custom field with `Email` label and `Email__c` name even though an existing property `Email` already exists?
> > Does salesforce enforce uniqueness on field labels in addition to field names?
> 
> Apparently not:
> 
> Screenshot 2023-02-09 at 10 28 53 AM

given that non-enforcement, imo it might be good to see the `__c` suffix if it's a commonly-understood format by customers using Supaglue. anybody else have thoughts on this?

khennes:
> imo it might be good to see the __c suffix if it's a commonly-understood format by customers

I think the expected behavior would be to see the display label rather than the name. If in the future we see that customers are actually creating identically labeled fields in their Salesforce instances, we can add some kind of text or indicator in the dropdown to disambiguate.

packages/nextjs/src/components/FieldMapping/FieldMapping.tsx
```diff
@@ -142,34 +147,31 @@ const FieldCollection = ({ appearance, syncConfig, sync }: FieldCollectionProps)
...
-        );
-      })}
+      {syncConfig.destination.schema.fields.map(({ label, name }, idx) => (
+        <div
+          key={idx}
+          css={styles.fieldWrapper}
+          className={classNames(appearance?.elements?.fieldWrapper, 'sg-fieldWrapper')}
+        >
+          <p css={styles.fieldName} className={classNames('sg-fieldName', appearance?.elements?.fieldName)}>
+            {label || name}
```
albertyfwu, line 157: nit: `label ?? name`?

albertyfwu (approved):

packages/nextjs/src/components/FieldMapping/FieldMapping.tsx
khennes: thanks will do!

albertyfwu (approved):

PR #96, opened 2023-02-09
feat(nextjs): Sync card top-level component, refactors

asdfryan:
* Create SyncConfigCard top-level component
* Move non-top-level to `elements` which is not exposed to developers
* Create `EmptyContent` primitive
* Various refactors
* Update docs

Light Mode
Screenshot 2023-02-09 at 3 19 39 PM

Dark Mode
Screenshot 2023-02-09 at 3 20 03 PM


tomkit:
- can we left-align the card instead of center-align the subcomponents?
- can you also update the next (docs/docs) and the next versioned_docs ?

albertyfwu:
> * can we left-align the card instead of center-align the subcomponents?
> * can you also update the next (docs/docs) and the next versioned_docs ?

don't we only need to update the `next` docs? I assume that when we cut a version, we'd copy the entire `next` docs into a new version

albertyfwu:
small thing but i personally think left-aligning the contents of the card looks better than center-aligning

asdfryan:
> * can we left-align the card instead of center-align the subcomponents?
> * can you also update the next (docs/docs) and the next versioned_docs ?

Re: left-alignt -- done

I think we should not be touching the versioned_docs.

albertyfwu (approved):

PR #97, opened 2023-02-09
docs: salesforce object config comment

albertyfwu:
Just a comment in code to explain difference between developer specifying an SF object vs expecting customer to choose one.


khennes (approved):
nice

PR #98, opened 2023-02-10
chore: check in markdown vscode settings

asdfryan:


lucasmarshall (changes requested):
Can we also add `"yzhang.markdown-all-in-one"` to extensions.json?

lucasmarshall (approved):

PR #101, opened 2023-02-10
docs: update docs for outbound sync

albertyfwu:


albertyfwu:
got offline approval from @george-xing

PR #103, opened 2023-02-10
refactor(sync): mirror sync code with config sdk interface

tomkit:
- main purpose was to get the `doSync` code in `do_sync.ts` to mirror the config sdk interface so you can express the sync workflow logic like:

```
const records = await sg.customers.sources.salesforce.X()
await sg.destinations.postgres.writeX(records);
```

- right now some of the context (sg, sync, syncConfig, syncRunId) are pass in at different levels of the fluent interface to have then be narrowed in scope [ question: is there a better way to do this? ]


apps/api/temporal/sdk/customer_integrations/index.ts
```diff
@@ -1,13 +1,30 @@
...
+    salesforce: SalesforceCustomerDestinationIntegration;
+  };
 };
 
-export const createCustomerIntegrations = (customerId: string) => {
+export const createCustomerIntegrations = (sync: Sync, syncConfig: SyncConfig, syncRunId: string) => {
   return {
-    salesforce: createSalesforce(customerId),
+    sources: {
+      salesforce: createSourceSalesforce(sync, syncConfig, syncRunId),
```
albertyfwu, line 22: nit: can we use the same ordering in names?

here `source` comes before `salesforce` while it's the opposite way for `SalesforceCustomerSourceIntegration`

apps/api/temporal/sdk/customer_integrations/salesforce.ts
```diff
@@ -102,4 +105,42 @@ export class SalesforceCustomerIntegration extends BaseCustomerIntegration {
...
 
-export const createSalesforce = (customerId: string) => new SalesforceCustomerIntegration(customerId);
+export class SalesforceCustomerSourceIntegration extends SalesforceCustomerIntegration {
+  public async bulkReadObjectType() {
+    const { sync, syncConfig } = this;
+    const fieldMapping = getMapping(sync, syncConfig);
+    const salesforceFields = [...Object.values(fieldMapping), 'SystemModstamp']; // TODO: Do not fetch SystemModstamp twice if already in mapping.
+    const salesforceFieldsString = salesforceFields.join(', ');
+
+    const salesforceObject = getSalesforceObject(syncConfig.source as SalesforceSource, sync);
```
albertyfwu, line 115: Ideally we can use generics on the classes so we don't need to do the casting `syncConfig.source as SalesforceSource` here. If we don't do it in this PR, maybe at least leave a TODO for better type safety?

This applies to all such places where we are casting.

apps/api/temporal/lib/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  return internalRecords.map((internalRecord) => {
+    const customerRecord: Record<string, unknown> = {};
+
+    Object.entries(mapping).forEach(([appField, customerField]) => {
+      customerRecord[customerField] = internalRecord[appField];
+    });
+
+    return customerRecord;
+  });
+}
```
albertyfwu, line 48: missing newline

apps/api/temporal/activities/lib/index.ts
```diff
@@ -0,0 +1,62 @@
+import { SalesforceCustomerIntegration, SyncConfig } from '../../../developer_config/entities';
```
albertyfwu, line 1: Can we move this file outside of the `activities` directory to avoid confusion that these are activities?

apps/api/temporal/sdk/internal_integrations/webhook.ts
```diff
@@ -55,3 +56,24 @@ export class WebhookInternalIntegration extends BaseInternalIntegration {
...
+export class WebhookDestinationInternalIntegration extends WebhookInternalIntegration {
+  public async sendRequests(records: any[]) {
+    const { sync, syncConfig, syncRunId } = this;
+    const fieldMapping = getMapping(sync, syncConfig);
+    const internalRecords = mapCustomerToInternalRecords(fieldMapping, records);
+    if (!internalRecords.length) {
+      throw new Error('No records to write');
+    }
+    for (const record of internalRecords) {
+      this.request(
```
albertyfwu, line 69: missing await

albertyfwu (commented):
Nice iteration on cleanup, though we still have more to do.

apps/api/temporal/sdk/customer_integrations/index.ts
tomkit: good catch, i missed going back to change it when i made a mental note

apps/api/temporal/activities/lib/index.ts
tomkit: yes

apps/api/temporal/sdk/internal_integrations/webhook.ts
tomkit: good catch

albertyfwu (approved):

PR #104, opened 2023-02-10
feat: migration to add custom_properties to syncs table

khennes:


apps/api/prisma/schema.prisma
```diff
@@ -32,12 +32,14 @@ model DeveloperConfig {
...
-  enabled        Boolean
-  syncConfigName String    @map("sync_config_name")
-  fieldMapping   Json?     @map("field_mapping")
+  id                String    @id @default(uuid())
+  customerId        String    @map("customer_id")
+  type              String
+  enabled           Boolean
+  syncConfigName    String    @map("sync_config_name")
+  fieldMapping      Json?     @map("field_mapping")
+  customProperties  Json?     @map("custom_properties")
```
albertyfwu, line 41: Why do we need a new column for this?

When we run the sync (`do_sync.ts`), we have access to both `SyncConfig` and `Sync`, or more specifically, we have access to both the `defaultFieldMapping` from the developer AND the `fieldMapping` from the customer.

Isn't that sufficient information to determine which key/values in `fieldMapping` are custom properties and which are not?

khennes: > Isn't that sufficient information to determine which key/values in fieldMapping are custom properties and which are not?

Well, we need to know about customer-defined properties in the FE as well, not just in the `do_sync` activity. We talked about doing it this way as an alternative, but 1) there is additional metadata about custom properties that we'll want/need to capture in the near future, such as data type, display name, etc., 2) making custom properties implicit (where we only know that a property is custom because it's not in the destination schema) is IMO a little too magical or hacky, and 3) similar to field mappings, custom properties should be defined once for a customer and be able to be reused across the application, not tied specifically to a single sync.

albertyfwu: Makes sense. I was also thinking if in the medium/long term we might want to stop defining `fieldMapping` as just `Record<string, string>` and something a bit more structured so it can also encapsulate the customer properties semantics, but that's probably something we can discuss when we actually do the work of ripping out `fieldMapping` and `customProperties` (to be sharable).

albertyfwu (approved):

apps/api/prisma/schema.prisma
khennes: Yeah I agree on both counts 👍

PR #105, opened 2023-02-10
feat: enable customers to create custom properties in the sync config

khennes:
Depends on #104

Adds a simple button to allow customers to create custom properties for the purposes of their SFDC integration. Note that these are application-side custom properties, not Salesforce custom fields.

For now, this is a fairly naive implementation that assumes that customer-defined properties aren't consumed by the application outside of the Salesforce syncs.

**Caveats**
1. Custom properties cannot be renamed
2. Custom properties cannot be reused across syncs, even for the same sObjects (same as field mappings). We're planning to refactor how syncs are modeled to enable field mappings, custom properties, etc. to be shared across syncs.

TO DO (follow-up):
- [ ] Actually write ingested Salesforce data to the custom properties
- [ ] Validate custom field when creating and display errors
- [ ] Fix annoying UI issue where the new custom property input lingers after the field mapping has been updated
- [ ] Only allow creating custom properties if enabled for that customer by the developer
- [ ] Update SG default theme to include form/input styles (FYI @asdfryan - I held off on this until you're back from OOO)

### DIFF
--- a/apps/api/routes/sync.ts
+++ b/apps/api/routes/sync.ts
@@ -2,7 +2,7 @@ import { Request, Response, Router } from 'express';
 import { getDependencyContainer } from '../dependency_container';
 import { BadRequestError, NotFoundError } from '../errors';
 import { errorMiddleware as posthogErrorMiddleware, middleware as posthogMiddleware } from '../lib/posthog';
-import { Sync, SyncCreateParams, SyncRun, SyncRunStatus } from '../syncs/entities';
+import { Sync, SyncCreateParams, SyncRun, SyncRunStatus, SyncUpdateParams } from '../syncs/entities';
 
 const { syncService, developerConfigService } = getDependencyContainer();
 
@@ -58,7 +58,7 @@ router.post(
 router.put(
   '/:syncId',
   posthogMiddleware('Update Sync'),
-  async (req: Request<{ syncId: string }, any, SyncCreateParams>, res: Response<Sync>) => {
+  async (req: Request<{ syncId: string }, any, SyncUpdateParams>, res: Response<Sync>) => {
     // TODO: Validate that `req.params.syncId` is equal to `req.body.id`.
     const developerConfig = await developerConfigService.getDeveloperConfig();
     const sync = await syncService.updateSync(req.params.syncId, req.body, developerConfig);
--- a/apps/api/syncs/entities/sync.ts
+++ b/apps/api/syncs/entities/sync.ts
@@ -194,8 +194,8 @@ type BaseSyncUpdateParams = {
   enabled: boolean;
   syncConfigName: string;
   fieldMapping?: Record<string, string>;
-  // Customer-defined fields that are not included in the developer's destination schema
-  customProperties?: Record<string, string>[];
+  // TODO: cast json column values to correct types when reading from db
+  customProperties?: Record<string, string>[]; // Customer-, rather than developer-defined
 };
 
 type BaseSyncCreateParams = BaseSyncUpdateParams & {
--- a/apps/api/syncs/entities/sync.ts
+++ b/apps/api/syncs/entities/sync.ts
@@ -194,6 +194,8 @@ type BaseSyncUpdateParams = {
   enabled: boolean;
   syncConfigName: string;
   fieldMapping?: Record<string, string>;
+  // Customer-defined fields that are not included in the developer's destination schema
+  customProperties?: Record<string, string>[];
 };
 
 type BaseSyncCreateParams = BaseSyncUpdateParams & {
--- a/apps/api/routes/field.ts
+++ b/apps/api/routes/field.ts
@@ -8,6 +8,11 @@ const { integrationService, developerConfigService } = getDependencyContainer();
 
 const router: Router = Router({ mergeParams: true });
 
+type SObjectField = {
+  name: string;
+  label: string;
+};
+
 router.get(
   '/',
   posthogMiddleware('Get Fields'),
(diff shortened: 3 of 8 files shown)


apps/api/syncs/entities/sync.ts
```diff
@@ -194,6 +194,8 @@ type BaseSyncUpdateParams = {
   enabled: boolean;
   syncConfigName: string;
   fieldMapping?: Record<string, string>;
+  // Customer-defined fields that are not included in the developer's destination schema
+  customProperties?: Record<string, string>[];
```
albertyfwu, line 198: Can we type this as `Field[]` instead of `Record<string, string>[]`? It looks like that is how it's typed on the FE.

apps/api/routes/sync.ts
```diff
@@ -58,7 +58,7 @@ router.post(
 router.put(
   '/:syncId',
   posthogMiddleware('Update Sync'),
-  async (req: Request<{ syncId: string }, any, SyncCreateParams>, res: Response<Sync>) => {
+  async (req: Request<{ syncId: string }, any, SyncUpdateParams>, res: Response<Sync>) => {
```
albertyfwu, line 61: nice catch

apps/api/syncs/entities/sync.ts
albertyfwu: Is it intentional that we aren't allow the customer to map from a salesforce custom field to a field? I agree that in most cases it's probably necessary, but I'm wondering if there can be collision in certain cases. Fine with punting that decision down the road too.

khennes: Nope, not intended, but what are you seeing that made you think that?

In the UI you can map a custom property to a Salesforce custom field:
Screenshot 2023-02-10 at 11 46 01 AM

albertyfwu: nvm you're right.

What about the comment about `customProperties?: Field[]`?

khennes: Oh I see, I think the comment is confusing because it's only valid for inbound syncs. (What will the destination schema look like for outbound syncs?) I'll delete the second half of the comment

khennes: I'm having some trouble typing this, ok if I punt for later?
```
api:lint-staged: syncs/services/sync_service.ts(28,12): error TS2352: Conversion of type 'import("/Users/katherinehennes/supaglue/apps/api/node_modules/.prisma/client/index").Sync' to type 'import("/Users/katherinehennes/supaglue/apps/api/syncs/entities/sync").Sync' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
api:lint-staged:   Type 'Sync' is not comparable to type 'OutboundSync'.
api:lint-staged:     Type 'Sync' is not comparable to type 'BaseSyncUpdateParams'.
api:lint-staged:       Types of property 'customProperties' are incompatible.
api:lint-staged:         Type 'JsonValue' is not comparable to type '{ name: string; label?: string | undefined; }[] | undefined'.
api:lint-staged:           Type 'JsonArray' is not comparable to type '{ name: string; label?: string | undefined; }[]'.
api:lint-staged:             Type 'JsonValue' is not comparable to type '{ name: string; label?: string | undefined; }'.
api:lint-staged:               Property 'name' is missing in type 'JsonValue[]' but required in type '{ name: string; label?: string | undefined; }'.
api:lint-staged: syncs/services/sync_service.ts(50,12): error TS2352: Conversion of type 'import("/Users/katherinehennes/supaglue/apps/api/node_modules/.prisma/client/index").Sync' to type 'import("/Users/katherinehennes/supaglue/apps/api/syncs/entities/sync").Sync' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
api:lint-staged: syncs/services/sync_service.ts(85,18): error TS2352: Conversion of type 'import("/Users/katherinehennes/supaglue/apps/api/node_modules/.prisma/client/index").Sync' to type 'import("/Users/katherinehennes/supaglue/apps/api/syncs/entities/sync").Sync' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
api:lint-staged: syncs/services/sync_service.ts(110,18): error TS2352: Conversion of type 'import("/Users/katherinehennes/supaglue/apps/api/node_modules/.prisma/client/index").Sync' to type 'import("/Users/katherinehennes/supaglue/apps/api/syncs/entities/sync").Sync' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
command (/Users/katherinehennes/supaglue/apps/api) yarn run lint-staged exited (1)
```

khennes: I guess for the json prisma type, all keys need to be optional.

khennes: https://github.com/prisma/prisma/issues/3219

albertyfwu: I think you just need to cast between `prisma.JsonValue` and `Field` in `sync_service.ts` whenever reading/writing to/from DB, but punting is fine for now too. Can you leave a TODO for that?

albertyfwu (approved):

apps/api/syncs/entities/sync.ts
khennes: sure, added 👍  ideally though we could just do the casting once, in one place.

albertyfwu (approved):

PR #110, opened 2023-02-10
feat: Support custom properties in inbound syncs to Postgres

khennes:
Write to custom properties when syncing objects from Salesforce to Postgres.

Does not include:
- [ ] Reading from custom properties when doing outbound syncs from postgres > salesforce
- [ ] Supporting custom properties for webhook destinations

### DIFF
--- a/apps/api/temporal/sdk/internal_integrations/postgres.ts
+++ b/apps/api/temporal/sdk/internal_integrations/postgres.ts
@@ -69,11 +69,21 @@ export class DestinationPostgresInternalIntegration extends PostgresInternalInte
 
     // TODO: What do we do if there are columns missing in the source?
     // TODO: Check that there are actually columns to sync over
-    const { upsertKey } = destination.config;
-    const dbFields = Object.keys(fieldMapping);
+    const { customPropertiesColumn, upsertKey } = destination.config;
+
+    const normalizedFields = destination.schema.fields.map((field) => field.name);
+    const dbFields = Object.keys(fieldMapping).filter((field) => normalizedFields.includes(field));
+    const customFields = Object.keys(fieldMapping).filter((field) => !normalizedFields.includes(field));
+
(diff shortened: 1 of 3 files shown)


apps/sample-app/prisma/schema.prisma
```diff
@@ -74,16 +74,16 @@ model SalesforceCreds {
...
+  id              Int      @id @default(autoincrement())
+  email           String   @db.VarChar(255)
+  customerId      String   @map("customer_id") @db.VarChar(255)
+  salesforceId    String?  @map("salesforce_id") @db.VarChar(255)
+  title           String?  @db.VarChar(255)
+  firstName       String?  @map("first_name") @db.VarChar(255)
+  lastName        String?  @map("last_name") @db.VarChar(255)
+  extraAttributes Json?    @map("extra_attributes")
+  createdAt       DateTime @default(now()) @map("created_at")
+  updatedAt       DateTime @updatedAt @map("updated_at")
```
khennes, line 86: Prettier changes

apps/api/temporal/sdk/internal_integrations/postgres.ts
```diff
@@ -86,7 +96,15 @@ export class DestinationPostgresInternalIntegration extends PostgresInternalInte
 
     // TODO: Do this in batches
     for (const mappedRecord of internalRecords) {
-      const values = dbFields.map((field) => mappedRecord[field] ?? '');
+      const values = [...normalizedFields].map((field) => mappedRecord[field] ?? '');
```
albertyfwu, line 99: why not just `const values = normalizedFields.map(...)`?

albertyfwu (approved):

apps/api/temporal/sdk/internal_integrations/postgres.ts
khennes: good catch, thanks! this was a set in an earlier version.

PR #117, opened 2023-02-13
feat(sdk): allow the defintion of realtime inbound syncs

lucasmarshall:


khennes (approved):
i think our config types should be cleaned up so that "type" ("realtime") is decoupled from direction (inbound/outbound), but not urgent.

albertyfwu (approved):
same comment as @khennes, but that data modeling decision should become clearer with time

PR #119, opened 2023-02-13
fix(nextjs): Fix nextjs package to support new DSL, use types package

asdfryan:
This PR:
* pulls the `Sync` typedef from `api` to `types` package
* starts using shared typedefs in the `nextjs` package
* Fixes components to support inbound + outbound syncs
* Reverts opportunities to a normal `inbound` webhook sync temporarily while the `realtime_inbound` is still getting built


packages/types/src/type_guards.ts
```diff
@@ -0,0 +1,24 @@
+import {
+  InboundSyncConfig,
+  InternalDestination,
+  OutboundSyncConfig,
+  PostgresDestination,
+  RealtimeInboundSyncConfig,
+  SyncConfig,
+} from '.';
+
+export const isRealtimeInboundSyncConfig = (syncConfig: SyncConfig): syncConfig is RealtimeInboundSyncConfig => {
```
albertyfwu, line 10: why do we need these? writing `syncConfig.name === 'realtime_inbound'` works for type narrowing

asdfryan: It does not work for me.

albertyfwu: discussed offline and seems like it works

albertyfwu (approved):

PR #121, opened 2023-02-14
feat: create entity_integration_configs db table migration

albertyfwu:
To support the new concept of `Entity`, we want to store `EntityIntegrationConfig` as a standalone object in the DB. This would consist of both `fieldMapping` and `customProperties`.

### DIFF
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -31,6 +31,22 @@ model DeveloperConfig {
   @@map("developer_configs")
 }
 
+model FieldMapping {
+  id              String   @id @default(uuid())
+  customerId      String   @map("customer_id")
+  entityName      String   @map("entity_name")
+  // Purposely using `integrationType` instead of foreign key `integrationId`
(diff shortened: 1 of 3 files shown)


apps/api/prisma/schema.prisma
```diff
@@ -31,6 +31,22 @@ model DeveloperConfig {
...
 }
 
+model FieldMapping {
+  id              String   @id @default(uuid())
+  customerId      String   @map("customer_id")
+  entityName      String   @map("entity_name")
+  // Purposely using `integrationType` instead of foreign key `integrationId`
+  // so that even if the customer disconnects (deletes) the Salesforce integration
+  // we keep the FieldMappings around (in case they re-connect later).
+  integrationType String   @map("integration_type")
```
khennes, line 41: just to confirm, this would be "salesforce," "hubspot," etc.?

khennes (approved):

apps/api/prisma/schema.prisma
albertyfwu: yes

khennes (approved):

PR #122, opened 2023-02-14
chore(integration card): add icon support. design pass on integration…

tomkit:
… card in sample-app

Also clarify application objects in tabs under App Object tab

Screenshot 2023-02-14 at 9 43 16 AM

Screenshot 2023-02-14 at 1 25 13 PM


apps/sample-app/styles/globals.css
```diff
@@ -1,3 +1,140 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+.sg-salesforceDisconnectButton {
```
asdfryan, line 5: Isn't this the wrong place to make this change? This should go as a default style

asdfryan (changes requested):
Can you post a screenshot of what the new card looks like on dark theme? (which is now part of the tutorial)

apps/sample-app/styles/globals.css
tomkit: I think there's not other way to do it now? Is there a way to change the connect and disconnect button outside of stylesheets?

tomkit:
> Can you post a screenshot of what the new card looks like on dark theme? (which is now part of the tutorial)
Screenshot 2023-02-14 at 1 15 26 PM

what's the best way to update the dark theme now ?

apps/sample-app/styles/globals.css
asdfryan: Yes. See all of the `styles.ts` files for all the elements/components/primitives.

packages/nextjs/src/components/IntegrationCard/IntegrationCard.tsx
```diff
@@ -35,6 +43,7 @@ const IntegrationCardInternal = ({ name, description, configurationUrl, appearan
 
   return (
     <Card className="sg-integrationCard" appearance={appearance}>
+      {UserIcon ? UserIcon : null}
```
lucasmarshall, line 46: `UserIcon` defaults to `null`, so the ternary shouldn't be needed.

apps/sample-app/pages/index.tsx
```diff
@@ -275,10 +275,9 @@ export default function Users({ contacts, count }: PageProps) {
         </header>
 
         <PageTabs className="mb-4" tabs={pageTabs} disabled={false} />
-        {activeTab === 'Contacts' && <ContactsTable initialUsers={contacts} initialTotalUsers={count} />}
-        {activeTab === 'Leads' && <LeadsTable initialUsers={[]} initialTotalUsers={0} />}
-        {activeTab === 'Accounts' && <AccountsTable initialUsers={[]} initialTotalUsers={0} />}
-        {activeTab === 'Opportunities' && <OpportunitiesTable initialUsers={[]} initialTotalUsers={0} />}
+        {activeTab === 'App Contacts' && <ContactsTable initialUsers={contacts} initialTotalUsers={count} />}
```
asdfryan, line 278: I think this will break the integration since right now we use the tab name to look for the sync config.

tomkit: Ok, it's aiming to clarify the app vs sfdc objects. I'll add in a label or something

lucasmarshall (approved):

asdfryan:
I think the tabs are broken on the Users page @tomkit

PR #130, opened 2023-02-14
fix(sample-app): no sticky header cell

lucasmarshall:


khennes (approved):

apps/sample-app/styles/globals.css
```diff
@@ -1,3 +1,10 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+/* daisyui wrongly only makes the first header cell sticky, and not the first row cells
+   we don't want any sticky
+ */
+.table th:first-child {
```
albertyfwu, line 8: cc @asdfryan is this where the CSS should go?

albertyfwu (approved):

apps/sample-app/styles/globals.css
asdfryan: Yeah this is not our component, so it's fine.

PR #132, opened 2023-02-14
fix: fix upsert query for custom fields

asdfryan:


apps/api/temporal/sdk/internal_integrations/postgres.ts
```diff
@@ -72,13 +72,10 @@ export class DestinationPostgresInternalIntegration extends PostgresInternalInte
...
 
     const normalizedFields = destination.schema.fields.map((field) => field.name);
-    const dbFields = Object.keys(fieldMapping).filter((field) => normalizedFields.includes(field));
     const customFields = Object.keys(fieldMapping).filter((field) => !normalizedFields.includes(field));
 
     const hasCustomFields = !!customPropertiesColumn && customFields.length;
-    if (hasCustomFields) {
-      dbFields.push(customPropertiesColumn);
-    }
+    const dbFields = hasCustomFields ? [...normalizedFields, customPropertiesColumn] : normalizedFields;
```
asdfryan, line 78: The problem is that `normalizedFields` (used to populated `values`) and `dbFields` (used to populate `dbFieldsString`) did not have the same sort order.

khennes (approved):

PR #138, opened 2023-02-15
fix(nextjs): fix xicon styling

asdfryan:
The real problem was that JSX was not properly interpreting the css. I also refactored the component styles to take into account theming (which is currently not necessary right now, but is convenient if in the future we choose to do so)


packages/nextjs/src/primitives/icons/xIcon.tsx
```diff
@@ -1,8 +1,9 @@
+/** @jsxImportSource @emotion/react */
```
asdfryan, line 1: this is the real fix

lucasmarshall (approved):

PR #146, opened 2023-02-15
feat: update custom property UX

khennes:
No functional change, this just updates the custom property UX such that we show a (disabled) salesforce field dropdown alongside the text input where you can add the name of your new custom property.

In order to allow selecting a salesforce field before naming a custom property, we'll need to refactor the FieldMapping component and allow custom properties to be updated as well as created and deleted. This raises some product questions about how we should model custom properties, whether they need IDs, etc., and whether supaglue config is the correct place for that metadata or whether it should actually live in the developer's application.

### DIFF
--- a/packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
+++ b/packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
@@ -14,7 +14,7 @@ import { ReactNode, useState } from 'react';
 import useSWR from 'swr';
 import useSWRMutation from 'swr/mutation';
 import { updateSync, useSalesforceIntegration } from '../../hooks/api';
-import { Select, SelectElements } from '../../primitives';
+import { Select, SelectElements, SelectOption } from '../../primitives';
 import { XIcon } from '../../primitives/icons';
 import { SupaglueProviderInternal } from '../../providers';
 import { useSupaglueContext } from '../../providers/SupaglueProvider';
@@ -214,6 +214,8 @@ const FieldCollection = ({ appearance, syncConfig, sync }: FieldCollectionProps)
             {label ?? name}
           </p>
 
+          <span>»</span>
+
           <Select
             appearance={appearance}
             disabled={!!upsertKey && name === upsertKey}
@@ -236,7 +238,13 @@ const FieldCollection = ({ appearance, syncConfig, sync }: FieldCollectionProps)
       ))}
 
       {isCreatingCustomProperty && (
-        <NewCustomPropertyForm appearance={appearance} onCreateCustomProperty={onCreateCustomProperty} />
+        <NewCustomPropertyForm
+          appearance={appearance}
+          onCreateCustomProperty={onCreateCustomProperty}
+          onRemoveCustomProperty={onRemoveCustomProperty}
+          onUpdateMappedField={onUpdateMappedField}
+          options={fields || []}
+        />
       )}
 
       {/* TODO: Add flag to enable/disable custom properties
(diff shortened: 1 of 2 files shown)


packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
```diff
@@ -255,29 +263,59 @@ const FieldCollection = ({ appearance, syncConfig, sync }: FieldCollectionProps)
...
+        value={customPropertyName}
+      />
+      <input css={styles.customPropertySubmitInput} type="submit" />
+
+      <span>»</span>
+
+      {/* TODO: Allow selecting the field before naming the custom property */}
+      <Select
+        appearance={appearance}
+        disabled
```
asdfryan, line 303: Should it become un-disabled once `customPropertyName` becomes non-empty?

packages/nextjs/src/elements/FieldMapping/styles.ts
```diff
@@ -90,10 +95,21 @@ const customPropertySubmitInput = _applyTheme((theme: SgTheme) =>
 const addCustomPropertyButton = _applyTheme((theme: SgTheme) =>
   css({
     backgroundColor: theme.colors.background,
-    border: 'none',
-    color: theme.colors.text,
-    marginTop: '1rem',
-    textAlign: 'start',
+    border: `1px solid ${indigo.indigo11}`,
+    borderRadius: '0.5rem',
+    color: indigo.indigo11,
```
asdfryan, line 100: This should probably be theme dependent?

packages/nextjs/src/elements/FieldMapping/styles.ts
```diff
@@ -90,10 +95,21 @@ const customPropertySubmitInput = _applyTheme((theme: SgTheme) =>
...
+    fontWeight: '500',
+    marginLeft: 'auto',
+    marginTop: '0.75rem',
+    padding: '0.375rem 0.75rem',
+    width: 'fit-content',
+    ':hover': {
+      backgroundColor: indigo.indigo2,
```
asdfryan, line 107: Same here

asdfryan (changes requested):
Do you mind posting screenshots of light theme and dark theme?

george-xing (approved):

packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
```diff
@@ -293,7 +331,7 @@ const AddCustomPropertyButton = ({
     {...props}
     type={undefined}
   >
-    + Add custom property
+    + Field
```
lucasmarshall, line 334: Is this the correct label?

packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
khennes: No, because we only save the new custom property on submit and on blur - a change event isn't sufficient.

packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
khennes: https://supaglue.slack.com/archives/C01GBBSL0JU/p1676422438151709?thread_ts=1676421312.128329&cid=C01GBBSL0JU

khennes:
> Do you mind posting screenshots of light theme and dark theme?

Screenshot 2023-02-15 at 11 07 39 AM

Screenshot 2023-02-15 at 11 13 49 AM

Screenshot 2023-02-15 at 11 14 00 AM

packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
asdfryan: I'm confused -- so this `Select` component is permanently disabled?

packages/nextjs/src/elements/FieldMapping/styles.ts
khennes: Sure, let's do that after cutting the new release though.

packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
khennes: yes, see PR description (and slack message)

packages/nextjs/src/elements/FieldMapping/styles.ts
asdfryan: I think the dark mode would look better (the text is barely legible) if we just force a white backgroundColor for now.

asdfryan (approved):

khennes:
Thanks @asdfryan! I'll address the other comments in a follow-up PR after we ship the new release.

PR #150, opened 2023-02-15
feat: add customPropertiesEnabled setting to sync configs

khennes:
Adds a separate config property to the `SyncConfig` that governs whether or not a customer is able to add custom fields to a field mapping. Note that for postgres integrations, a `customPropertiesColumn` is also required.

Also fixes an issue where `destination.schema.fields` is out of sync with the `defaultFieldMapping`, so that fields specified in the latter but not the former do not appear in the customer's field mapping.

### DIFF
--- a/packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
+++ b/packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
@@ -99,8 +99,8 @@ const FieldCollection = ({ appearance, syncConfig, sync }: FieldCollectionProps)
 
   // Use the customer-defined field mapping if it exists; default to the values supplied by the developer
   const initialFieldMapping: CustomerFieldMapping = {};
-  (syncConfig.defaultFieldMapping || []).map(({ name, field }) => {
-    initialFieldMapping[name] = field;
+  ((syncConfig.destination as PostgresDestination).schema.fields || []).map(({ name }) => {
+    initialFieldMapping[name] = syncConfig.defaultFieldMapping?.find((field) => field.name === name)?.field || '';
   });
 
   if (sync.fieldMapping) {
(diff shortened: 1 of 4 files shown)


lucasmarshall (approved):

packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
```diff
@@ -29,10 +29,18 @@ const getSchema = (syncConfig: SyncConfig): Schema => {
 };
 
 const customPropertiesEnabled = (syncConfig: SyncConfig): boolean => {
-  if (syncConfig.type === 'outbound') {
-    return !!syncConfig.source.config.customPropertiesColumn;
+  if (syncConfig.type === 'inbound' && syncConfig.destination.type === 'postgres') {
+    return Boolean(
+      (syncConfig.destination as PostgresDestination).config.customPropertiesColumn &&
```
asdfryan, line 34: PSA: We should not be casting like this anymore and instead be using type guards to get the correct config (this is why outbound syncs are not working in the UI).

packages/nextjs/src/elements/FieldMapping/FieldMapping.tsx
```diff
@@ -91,8 +99,8 @@ const FieldCollection = ({ appearance, syncConfig, sync }: FieldCollectionProps)
 
   // Use the customer-defined field mapping if it exists; default to the values supplied by the developer
   const initialFieldMapping: CustomerFieldMapping = {};
-  (syncConfig.defaultFieldMapping || []).map(({ name, field }) => {
-    initialFieldMapping[name] = field;
+  ((syncConfig.destination as PostgresDestination).schema.fields || []).map(({ name }) => {
```
asdfryan, line 102: Same here

PR #152, opened 2023-02-16
chore: add more info about custom properties to docs

khennes:


albertyfwu (approved):

asdfryan (commented):
I thought we were supposed to be updating the `next` docs? unless the intent is for this to be part of the current release?

albertyfwu:
> I thought we were supposed to be updating the `next` docs? unless the intent is for this to be part of the current release?

True; and even if we want to update for current release, we should then also update `next`, since this is relevant for both sets of docs.

khennes:
good catch, thanks!

PR #168, opened 2023-02-28
fix: fix various analytics issues

khennes:


docker-compose.override.dev.yml
```diff
@@ -4,7 +4,6 @@ services:
     environment:
       - NODE_ENV=development
```
lucasmarshall, line 5: Probably don't want to check this in

lucasmarshall (changes requested):
Need to keep the analytics disabled for the dev override

apps/api/index.ts
```diff
@@ -45,6 +45,8 @@ app.use(
   })
 );
 
+app.use(posthogMiddleware);
```
lucasmarshall, line 48: Doesn't this log success always, even for errors?

khennes:
> Need to keep the analytics disabled for the dev override

Why do we want that?

apps/api/index.ts
khennes: Yes, it's more of an "attempt" than a success. I couldn't figure out how to emit an event for a successful request.

khennes:
> > Need to keep the analytics disabled for the dev override
> 
> Why do we want that?

For posterity: we'll keep analytics enabled for the dev override so that we have visibility into external folks who want to develop against supaglue.

apps/api/index.ts
khennes: We'll rename the result type to `attempt` for now and figure out the middleware later.

PR #182, opened 2023-03-01
fix: docs builds

lucasmarshall:


turbo.json
```diff
@@ -6,17 +6,6 @@
...
-        "../../openapi/**/openapi.bundle.json",
-        "blog/**/*",
-        "docs/**/*",
-        "docusaurus.config.js",
-        "sidebars.js",
-        "package.json",
-        "tsconfig.json",
-        "src/**/*",
-        "static/**/*"
-      ],
```
lucasmarshall, line 19: This is the fix. Turbo will skip cache if anything in the docs workspace changes with the new config

albertyfwu (approved):

turbo.json
tomkit: Thanks!

PR #187, opened 2023-03-02
fix: db issues

lucasmarshall:
There are some future changes I want to make here to simplify the Dockerfiles, but this works for now


docker-compose.yml
```diff
@@ -60,17 +60,15 @@ services:
       - temporalitedata:/data
 
   init:
-    image: node:18
+    image: supaglue/init
```
tomkit, line 63: do we need an init in override.yml?

lucasmarshall: No.

albertyfwu (approved):

PR #191, opened 2023-03-02
chore: create sync history table

lucasmarshall:


packages/db/prisma/schema.prisma
```diff
@@ -166,3 +167,18 @@ model CrmOpportunity {
   @@unique([connectionId, remoteId])
   @@map("crm_opportunities")
 }
+
+model SyncHistory {
+  id             String     @id @default(uuid())
+  // contact, lead, account, opportunity, etc.
+  object         String
```
albertyfwu, line 174: we should standardize on `object` or `model`. i think we're using both in the code...

packages/db/prisma/schema.prisma
```diff
@@ -166,3 +167,18 @@ model CrmOpportunity {
...
+  // success | error
+  result         String
+  startTimestamp DateTime   @default(now()) @map("start_timestamp")
```
albertyfwu, line 177: why do we want default here?

packages/db/prisma/schema.prisma
```diff
@@ -33,19 +33,20 @@ model Integration {
...
   credentials   Json // {type, access_token, refresh_token, expires_at, raw}
-  customer      Customer    @relation(fields: [customerId], references: [id], onDelete: Cascade)
-  customerId    String      @map("customer_id")
-  createdAt     DateTime    @default(now()) @map("created_at")
-  updatedAt     DateTime    @updatedAt @map("updated_at")
+  customer      Customer      @relation(fields: [customerId], references: [id], onDelete: Cascade)
+  customerId    String        @map("customer_id")
+  createdAt     DateTime      @default(now()) @map("created_at")
+  updatedAt     DateTime      @updatedAt @map("updated_at")
+  SyncHistory   SyncHistory[]
```
albertyfwu, line 49: lowercase? 

also, do we need to call it `syncHistoryList`? For connections, we're calling it `connections: Connection[]`

lucasmarshall: Will update. This is the way the prisma extension for vscode auto-created this.

packages/db/prisma/schema.prisma
lucasmarshall: I guess we dont

packages/db/prisma/schema.prisma
lucasmarshall: will switch to model

albertyfwu (approved):

PR #196, opened 2023-03-03
fix: sf bulk 2.0

albertyfwu:
Somehow jsforce was hanging on bulk 2.0 query. didn't debug and just rewrote it with `fetch`.

This does not yet deal with refreshing access token.

### DIFF
--- a/packages/core/remotes/crm/salesforce/index.ts
+++ b/packages/core/remotes/crm/salesforce/index.ts
@@ -1,6 +1,9 @@
+// fetch methods / types not present in @types/node yet
+// https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924
+/// <reference lib="dom" />
+
 import { CsvError, Info, parse } from 'csv-parse';
 import * as jsforce from 'jsforce';
-import fetch, { type Response } from 'node-fetch';
 import { AccountCreateParams, RemoteAccount, RemoteAccountUpdateParams } from '../../../types/account';
 import { CRMConnection } from '../../../types/connection';
 import { RemoteContact, RemoteContactCreateParams, RemoteContactUpdateParams } from '../../../types/contact';
@@ -26,9 +29,6 @@ import {
   toSalesforceOpportunityCreateParams,
   toSalesforceOpportunityUpdateParams,
 } from './mappers';
-
-// TODO: data-drive
-const BULK2_POLL_TIMEOUT = 240000; // 4 minutes
 
 const propertiesToFetch = {
   account: [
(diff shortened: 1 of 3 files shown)


packages/core/remotes/crm/salesforce/index.ts
```diff
@@ -1,4 +1,6 @@
+import { CsvError, Info, parse } from 'csv-parse';
 import * as jsforce from 'jsforce';
+import fetch, { type Response } from 'node-fetch';
```
lucasmarshall, line 3: Where is this being added as a dependency? I don't see it in the `package.json` above. Also, we could just use the [built-in node version](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch) of fetch since we are on node 18

packages/core/remotes/crm/salesforce/index.ts
```diff
@@ -150,13 +195,112 @@ class SalesforceClient extends CrmRemoteClientEventEmitter implements CrmRemoteC
...
+    throw new Error('bulk 2.0 job polling timed out');
+  }
+
+  async #getBulk2QueryJobResponse(jobId: string, locator?: string): Promise<Response> {
+    const params = new URLSearchParams();
+    if (locator) {
+      params.set('locator', locator);
+    }
+
+    const response = await fetch(`${this.#instanceUrl}/services/data/v57.0/jobs/query/${jobId}/results?${params}`, {
```
lucasmarshall, line 265: Same as above

packages/core/remotes/crm/salesforce/index.ts
```diff
@@ -150,13 +195,112 @@ class SalesforceClient extends CrmRemoteClientEventEmitter implements CrmRemoteC
...
+    if (response.status !== 200) {
+      throw new Error(`Status code ${response.status} when submitting salesforce bulk 2.0 query`);
+    }
+
+    return await response.json();
+  }
+
+  async #pollBulk2QueryJob(jobId: string): Promise<void> {
+    const poll = async (): Promise<SalesforceBulk2QueryJob> => {
+      const response = await fetch(`${this.#instanceUrl}/services/data/v57.0/jobs/query/${jobId}`, {
```
lucasmarshall, line 220: Same as above

packages/core/remotes/crm/salesforce/index.ts
```diff
@@ -150,13 +195,112 @@ class SalesforceClient extends CrmRemoteClientEventEmitter implements CrmRemoteC
     });
   }
 
+  async #submitBulk2QueryJob(soql: string): Promise<SalesforceBulk2QueryJob> {
+    const response = await fetch(`${this.#instanceUrl}/services/data/v57.0/jobs/query`, {
```
lucasmarshall, line 199: Do we want the version to be config-driven here?

albertyfwu: maybe eventually, but this seems good for now, especially because changing a version may not just be a matter of updating it in the URL, but changing processing code, since assumptions around data schema could change.

khennes: nit: use a constant for the version #?

lucasmarshall (approved):

PR #197, opened 2023-03-04
feat: write to sync history

lucasmarshall:


packages/sync-workflows/workflows/run_syncs.ts
```diff
@@ -16,6 +18,24 @@ export type RunSyncsArgs = {
...
+      } else {
+        await logSyncFinish({
+          historyId: historyIds[idx],
+          status: 'FAILURE',
+          errorMessage: result.reason.message ?? 'Unknown error',
+        });
+      }
+    })
+  );
   await populateAssociations({ connectionId });
```
albertyfwu, line 40: This is probably beyond the scope of this PR, but we should probably leave a note somewhere that we need to address the "sync" abstraction so that we can properly account for the fact that associations are not done yet by the time that `logSyncFinish` is recorded.

packages/sync-workflows/activities/log_sync_finish.ts
```diff
@@ -0,0 +1,23 @@
...
+  return async function logSyncFinish({
+    historyId,
+    status,
+    errorMessage,
+  }: {
+    historyId: number;
+    status: SyncHistoryStatus;
+    errorMessage?: string;
+  }) {
+    await syncHistoryService.update({
```
albertyfwu, line 14: Could we move this business logic inside `SyncHistoryService` instead instead of exposing a generic `update` method? It doesn't seem like there will be any external trigger of these methods; they're all internal uses.

packages/db/prisma/schema.prisma
```diff
@@ -169,11 +169,11 @@ model CrmOpportunity {
 }
 
 model SyncHistory {
-  id             String     @id @default(uuid())
+  id             Int        @id @default(autoincrement())
   // contact, lead, account, opportunity, etc.
   model          String
   // success | error
```
albertyfwu, line 175: update comment? I believe `IN_PROGRESS` is a valid status

packages/core/types/sync_history.ts
```diff
@@ -0,0 +1,15 @@
...
+export type SyncHistoryStatus = 'SUCCESS' | 'FAILURE' | 'IN_PROGRESS';
+
+export type SyncHistory = {
+  id: number;
+  model: string;
+  status: SyncHistoryStatus;
+  errorMessage: string | null;
+  startTimestamp: Date;
+  endTimestamp: Date | null;
+  connection: Connection;
```
albertyfwu, line 12: maybe we can address in another PR, but doesn't this include the credentials too? Is that necessary for a consumer of `SyncHistory`?

packages/core/services/sync_history_service.ts
```diff
@@ -0,0 +1,86 @@
...
+    paginationParams,
+    model,
+  }: {
+    connectionId: string;
+    paginationParams: PaginationParams;
+    model?: string;
+  }): Promise<PaginatedResult<SyncHistory>> {
+    const { page_size, cursor } = paginationParams;
+    const pageSize = page_size ? parseInt(page_size) : undefined;
+    // TODO casting to any here because prisma isn't typing the model with the included connection correctly
```
albertyfwu, line 62: is the todo to make it generic later?

albertyfwu (approved):

packages/core/services/sync_history_service.ts
lucasmarshall: The TODO is to revisit if/when prisma fixes this typing issue

packages/core/types/sync_history.ts
lucasmarshall: Yeah, we can probably narrow the fields returned here to only include what a consumer may need. Will do.

packages/core/mappers/sync_history.ts
```diff
@@ -0,0 +1,23 @@
...
+  connection,
+}: SyncHistoryModel & { connection: Connection }): SyncHistory => {
+  return {
+    id,
+    model,
+    status: status as SyncHistoryStatus,
+    errorMessage,
+    startTimestamp,
+    endTimestamp,
+    connection: fromConnectionModel(connection),
```
albertyfwu, line 21: does this still return the `credentials`? Additionally, why does the caller of `syncHistoryModel` need the connection object?

lucasmarshall: I'll remove it for now

albertyfwu (approved):

PR #204, opened 2023-03-07
feat: endpoint for sync-history

lucasmarshall:
Also restructures the openapi a bit to allow for customer API docs, fixes some error handling


apps/api/routes/customer/integration/connection/index.ts
```diff
@@ -28,5 +30,8 @@ export default function init(app: Router): void {
     return res.status(200).send(connection);
   });
 
+  syncInfo(connectionRouter);
```
albertyfwu, line 33: Given that we pass in the customer id in the headers, isn't it enough to look up the sync info? Why do we want to make the sync info and sync history endpoints nested under `/connections`?

lucasmarshall: We don't pass them in the headers for the customer api

apps/api/middleware/connection.ts
```diff
@@ -6,7 +6,7 @@ const { connectionService, integrationService } = getDependencyContainer();
 
 export async function connectionMiddleware(req: any, res: Response, next: NextFunction) {
   req.sg = {
-    connectionId: req.params.connectionId,
+    connectionId: req.params.connectionId ?? req.params.connection_id,
```
tomkit, line 9: how come we need to support both?

apps/api/routes/customer/integration/connection/index.ts
tomkit: I agree sync-info sync-history should be a top-level API and the developer should pass in params via headers rather than nested under CRM APIs

apps/api/routes/customer/integration/connection/sync_history.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export default function init(app: Router) {
+  app.get(
+    '/sync-history',
+    async (
+      req: Request<GetSyncHistoryPathParams, GetSyncHistoryResponse, GetSyncHistoryRequest, GetSyncHistoryQueryParams>,
+      res: Response<GetSyncHistoryResponse>
+    ) => {
+      const { next, previous, results } = await syncHistoryService.list({
+        connectionId: req.params.connection_id,
```
tomkit, line 21: how did you get req.params to work? if the param is not in the immediate router URI?

lucasmarshall: It is in the immediate router.

lucasmarshall: Also, if you want, you can have params merged from parent routers.

apps/api/middleware/connection.ts
lucasmarshall: The latter matches the openapi spec, but the former is being used in endpoints that don't have an openapi spec

apps/api/routes/customer/integration/connection/index.ts
lucasmarshall: So you are saying this should be moved out of `/customers` to someplace else?

sync-info was in the CRM API before, but it didn't make sense there. Should we move both of these to a sync API?

apps/api/middleware/connection.ts
tomkit: Ok, (my bad for using it in express), I think we should just have express route.use use `connection_id` rather than `connectionId` then

apps/api/routes/customer/integration/connection/index.ts
tomkit: For posterity we walked this back

asdfryan (approved):

apps/api/middleware/connection.ts
```diff
@@ -6,7 +6,7 @@ const { connectionService, integrationService } = getDependencyContainer();
 
 export async function connectionMiddleware(req: any, res: Response, next: NextFunction) {
   req.sg = {
-    connectionId: req.params.connectionId,
```
tomkit, line 9: does the customer middleware also need to be updated?

apps/api/routes/crm/v1/sync_history.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export default function init(app: Router) {
+  app.get(
+    '/sync-history',
+    async (
+      req: Request<GetSyncHistoryPathParams, GetSyncHistoryResponse, GetSyncHistoryRequest, GetSyncHistoryQueryParams>,
+      res: Response<GetSyncHistoryResponse>
+    ) => {
+      const { next, previous, results } = await syncHistoryService.list({
+        connectionId: req.customerConnection.id,
```
tomkit, line 21: nit: at some point we probably should place all of our hydrated objects under a namespaced key, e.g. `sg`

apps/api/routes/customer/integration/index.ts
```diff
@@ -35,5 +35,5 @@ export default function init(app: Router): void {
 
   connection(integrationRouter);
 
-  app.use('/integrations/:integrationId', integrationMiddleware, integrationRouter);
```
tomkit, line 38: does the integration middleware also need to be updated?

apps/api/middleware/connection.ts
lucasmarshall: Good catch

apps/api/routes/crm/v1/sync_history.ts
lucasmarshall: Agreed, outside the scope of this PR

tomkit (approved):

PR #210, opened 2023-03-07
fix: fix non-crm router. add mgmt/v1/ routes

tomkit:


apps/api/routes/index.ts
```diff
@@ -1,14 +1,14 @@
...
 export default function initRoutes(app: Router): void {
   oauth(app);
+  mgmt(app);
 
-  customer(app);
-
-  app.use(connectionHeaderMiddleware);
-  crm(app);
+  const crmRouter = Router();
+  crmRouter.use(connectionHeaderMiddleware);
```
tomkit, line 12: previously connectionHeaderMiddleware was being applied for all routes under `app`

lucasmarshall (approved):

PR #213, opened 2023-03-08
chore: config for deploying to aws

lucasmarshall:
Init doesn't work, but this is a start, just wanted to get eyes on it


docker-compose.yml
```diff
@@ -5,7 +5,6 @@ services:
       - '5432:5432'
     volumes:
       - pgdata:/var/lib/postgresql/data
-      - ./docker/postgres/docker-entrypoint-initdb.d:/docker-entrypoint-initdb.d
```
lucasmarshall, line 8: Unused

albertyfwu (dismissed):

lucasmarshall:
> I think we can keep this on a branch for a week or two until we need to try ECS to keep our docker compose files simpler on main, thoughts?

Concerned about merge conflicts if we do that

README.md
```diff
@@ -55,6 +64,82 @@ Continue on to the docs to go through our [quickstart](https://docs.supaglue.com
 
 We are currently in Public Alpha. Watch "releases" of this repo to be notified of significant updates (as minor semver releases).
 
+## Self hosting
```
tomkit, line 67: Maybe move this to DEVELOPMENT.md

lucasmarshall: It isn't development related. This is deployment.

tomkit: Can we move it into the docs then? I think it's too much real estate for this on our main readme.md

PR #221, opened 2023-03-08
feat: send webhook upon successful / errored connection

asdfryan:


apps/api/services/connection_writer_service.ts
```diff
@@ -58,23 +66,40 @@ export class ConnectionWriterService {
...
+        connection.id,
+        integration.config.sync.periodMs ?? FIFTEEN_MINUTES_MS
+      );
+      return fromConnectionModel(connection);
+    } catch (e) {
+      errored = true;
+      throw e;
+    } finally {
+      if (application.config.webhook) {
+        await sendWebhookPayload(
```
tomkit, line 97: Should we just do fire-and-forget now for these notification webhooks? We can look into queueing and making them lossless down the road (maybe take a look at svix at that point too)?

asdfryan: Are you suggesting we don't await?

albertyfwu: until we do queueing, not awaiting might be better, since it doesn't return an error for the actual main purpose of this function, which is to create a connection. indeed, even if the webhook fails to send here, the connection has been created successfully in the DB.

don't have a strong opinion though.

packages/core/services/application_service.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!application) {
+      throw new NotFoundError(`Can't find application with id: ${id}`);
+    }
+    return fromApplicationModel(application);
+  }
+
+  // TODO: paginate
+  public async list(): Promise<Application[]> {
+    const applications = await this.#prisma.application.findMany();
+    return applications.map((application) => fromApplicationModel(application));
```
albertyfwu, line 26: nit: can just do `return applications.map(fromApplicationModel)`

packages/core/services/application_service.ts
```diff
@@ -0,0 +1,54 @@
...
+    const updatedApplication = await this.#prisma.application.update({
+      where: { id },
+      data: {
+        ...application,
+      },
+    });
+    return fromApplicationModel(updatedApplication);
+  }
+
+  public async delete(id: string): Promise<Application> {
```
albertyfwu, line 48: is it necessary to return the object on delete?

albertyfwu (approved):

packages/core/services/application_service.ts
asdfryan: No, but that is the convention in our services thus far.

apps/api/services/connection_writer_service.ts
asdfryan: I made `sendWebhookPayload` best effort (and refactored logger)

albertyfwu (approved):
can you leave TODO for casting pino logger?

albertyfwu (approved):

PR #227, opened 2023-03-09
chore: customer expansion to connections

tomkit:


packages/core/mappers/customer.ts
```diff
@@ -1,10 +1,18 @@
...
+  const expandConnections = expandedAssociations.includes('connections');
   return {
     id,
     applicationId,
     createdAt,
     updatedAt,
-  } as Customer;
+    connections:
+      expandConnections && connections ? connections.map((connection) => fromConnectionModel(connection)) : undefined,
+  } as CustomerExpanded;
```
albertyfwu, line 17: Can we avoid type assertions?

albertyfwu: Also, if we return `CustomerExpanded` but the return type of this function is `Customer`, then we don't get any more type-checking.

packages/core/mappers/customer.ts
```diff
@@ -1,10 +1,18 @@
-import type { Customer, Customer as CustomerModel } from '@supaglue/db';
+import type { Customer } from '@supaglue/db';
+import { CustomerExpanded } from '../types/customer';
+import { fromConnectionModel } from './connection';
 
-export const fromCustomerModel = ({ id, applicationId, createdAt, updatedAt }: CustomerModel): Customer => {
+export const fromCustomerModel = (
+  { id, applicationId, createdAt, updatedAt, connections }: CustomerExpanded,
```
asdfryan, line 6: This method signature doesn't make sense to me. Based on the name it looks like you are converting from prisma model to the entity model, but `CustomerExpanded` is not the prisma model.

packages/core/mappers/customer.ts
```diff
@@ -1,10 +1,18 @@
...
+  expandedAssociations: string[] = []
```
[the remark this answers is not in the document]
tomkit: The same thing in our application, in Prisma, expanded relations

packages/core/mappers/customer.ts
tomkit: +1, imported the wrong Customer

packages/core/mappers/customer.ts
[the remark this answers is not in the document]
asdfryan: What I mean is associations is a product terminology -- I think this method should just pass a boolean instead for whether or not to expand connections.

packages/core/mappers/customer.ts
tomkit: Good catch

packages/core/mappers/customer.ts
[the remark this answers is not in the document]
tomkit: We'll assume we only need 1-hop joins now and have all relations joined if `includeRelations` is set to true

asdfryan (approved):

PR #231, opened 2023-03-09
chore: pipeline writes into Postgres to avoid OOM

albertyfwu:
Changed interfaces for `CrmRemoteClient` for `listAccounts`, `listContacts`, ... to return `Promise<Readable>` instead to pipeline writes directly into the DB.

The code is a bit complex right now, and we should take some time to simplify it.


packages/core/remotes/crm/salesforce/index.ts
```diff
@@ -296,32 +297,64 @@ class SalesforceClient extends CrmRemoteClientEventEmitter implements CrmRemoteC
...
 
-  public async listAccounts(): Promise<RemoteAccount[]> {
+  private async listCommonModelRecords(soql: string, mapper: (record: Record<string, any>) => any): Promise<Readable> {
+    return pipeline(
+      await this.#getBulk2QueryJobResults(soql),
+      new Transform({
+        objectMode: true,
+        transform: (chunk, encoding, callback) => {
+          try {
+            callback(null, mapper(chunk));
```
tomkit, line 341: 🚀

packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -187,15 +204,31 @@ class HubSpotClient extends CrmRemoteClientEventEmitter implements CrmRemoteClie
...
-    let after = undefined;
-    const remoteOpportunities = [];
-    do {
-      const currResult: HubspotPaginatedDeals = await this.listOpportunitiesImpl(after);
-      after = currResult.paging?.next?.after;
-      remoteOpportunities.push(...currResult.results.map(fromHubSpotDealToRemoteOpportunity));
-    } while (after);
-    return remoteOpportunities;
+  public async listOpportunities(): Promise<Readable> {
+    const passThrough = new PassThrough({ objectMode: true });
```
asdfryan, line 208: seems like there is a way to DRY this, but can be done later.

asdfryan (dismissed):

packages/core/remotes/crm/hubspot/index.ts
albertyfwu: yes i tried but it was not very elegant. we can revisit later.

asdfryan (approved):

PR #241, opened 2023-03-10
fix: broken openapi specs/validation middleware

albertyfwu:
Most of the issues here are due to ajv having a bug and interpreting `id: <uuid>` as schema id even when they're used in examples. See https://github.com/stoplightio/spectral/issues/2081

Also, we seem to be mixing openapi 3.0.3 and 3.1.0 syntax for examples, which we should clean up at some point.

### DIFF
--- a/apps/api/routes/mgmt/v1/index.ts
+++ b/apps/api/routes/mgmt/v1/index.ts
@@ -1,9 +1,12 @@
+import { openapiMiddleware } from '@/middleware/openapi';
 import { Router } from 'express';
 import customer from './customer';
 import integration from './integration';
 
 export default function init(app: Router): void {
   const v1Router = Router();
+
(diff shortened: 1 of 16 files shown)


apps/api/routes/mgmt/v1/index.ts
```diff
@@ -1,10 +1,13 @@
+import { openapiMiddleware } from '@/middleware/openapi';
 import { Router } from 'express';
 import customer from './customer';
 import integration from './integration';
 
 export default function init(app: Router): void {
   const v1Router = Router();
 
+  v1Router.use(openapiMiddleware('mgmt'));
```
albertyfwu, line 9: this was the main issue that was not caused by `ajv` issues

lucasmarshall (approved):

PR #250, opened 2023-03-11
fix: downgrade hubspot client to fix token refresh and add temporal heartbeating

albertyfwu:
We can get back to a more modern version of hubspot-api-nodejs when https://github.com/HubSpot/hubspot-api-nodejs/issues/341 is fixed.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,4 +1,4 @@
-nmHoistingLimits: workspaces
+nmHoistingLimits: none
 
 nodeLinker: node-modules
 
--- a/packages/core/remotes/crm/hubspot/index.ts
+++ b/packages/core/remotes/crm/hubspot/index.ts
@@ -34,7 +34,7 @@ const ASYNC_RETRY_OPTIONS = {
   forever: true,
   factor: 2,
   minTimeout: 1000,
-  maxTimeout: 60 * 6000,
+  maxTimeout: 60 * 1000,
 };
 
 const propertiesToFetch = {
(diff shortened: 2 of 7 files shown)


.yarnrc.yml
```diff
@@ -1,4 +1,4 @@
-nmHoistingLimits: workspaces
+nmHoistingLimits: none
```
albertyfwu, line 1: without this, temporal heartbeating (calling `Context.current()`) was not working because async local storage was created / accessed from different copies of the same library

packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -34,7 +34,7 @@ const ASYNC_RETRY_OPTIONS = {
   forever: true,
   factor: 2,
   minTimeout: 1000,
-  maxTimeout: 60 * 6000,
```
albertyfwu, line 37: 6000 looks like a mistake

packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -112,7 +112,6 @@ class HubSpotClient extends CrmRemoteClientEventEmitter implements CrmRemoteClie
     const { accessToken } = credentials;
     this.#client = new Client({
       accessToken,
-      numberOfApiCallRetries: 1,
```
albertyfwu, line 115: 8.3.2 doesn't support this

packages/db/package.json
```diff
@@ -27,5 +27,8 @@
   ],
   "prisma": {
     "seed": "tsx prisma/seed.ts"
+  },
+  "installConfig": {
+    "hoistingLimits": "workspaces"
```
albertyfwu, line 32: do not hoist beyond workspace for `db` because prisma generate doesn't work well with that

.yarnrc.yml
albertyfwu: cc @lucasmarshall in case this is anticipated to break anything in the build process

PR #255, opened 2023-03-13
fix: convert customer create/update to be upsert

tomkit:


albertyfwu (commented):
Why do we want to make it upsert?

packages/db/prisma/schema.prisma
```diff
@@ -23,15 +23,17 @@ model Application {
...
-  createdAt     DateTime     @default(now()) @map("created_at")
-  updatedAt     DateTime     @updatedAt @map("updated_at")
-
+  id                 String       @id @default(uuid())
+  externalIdentifier String       @map("external_identifier")
+  applicationId      String       @map("application_id")
+  application        Application  @relation(fields: [applicationId], references: [id], onDelete: Cascade)
+  connections        Connection[]
+  name               String       @db.VarChar(255)
+  email              String       @db.VarChar(320)
```
asdfryan, line 32: Why 320?

asdfryan: Alternatively, I think it's fine to remove the `@db.VarChar()` decorator and just treat as text

albertyfwu (approved):

packages/db/prisma/schema.prisma
tomkit: posterity: email spec limits email length to 320
decision is to drop all varchar limits to just be text

tomkit:
> Is there any reason for a globally-unique `customerId`? There seems to be no reason to access it solo. Usually you would have an `applicationId` in context.

Solving for onboarding flow for now, SUP1-113 to decide what combination of external_identifier vs id we want to support

albertyfwu (approved):

PR #257, opened 2023-03-13
feat: user common model endpoints, service, type, and mapper

asdfryan:
TODO:
- open API spec + typing
- hubspot/salesforce implementation
- owner associations


apps/api/routes/crm/v1/user.ts
```diff
@@ -0,0 +1,44 @@
...
+import { camelcaseKeys } from '@/lib/camelcase';
+import { snakecaseKeys } from '@/lib/snakecase';
+import { Request, Response, Router } from 'express';
+
+const { userService } = getDependencyContainer();
+
+export default function init(app: Router): void {
+  const router = Router();
+
+  router.get('/', async (req: Request, res: Response) => {
```
albertyfwu, line 11: can we use the actual types to have type-checking?

packages/core/types/crm/lead.ts
```diff
@@ -1,5 +1,5 @@
 import type { CrmAccount, CrmContact, CrmLead } from '@supaglue/db';
-import type { Account, Address, Contact, EmailAddress, PhoneNumber } from '.';
+import type { Account, Address, Contact, EmailAddress, PhoneNumber } from '../index';
```
albertyfwu, line 2: just `..` is enough

albertyfwu (approved):

apps/api/routes/crm/v1/user.ts
asdfryan: Yes I will, but I haven't done the open API spec yet so we don't have the types (next PR).

packages/core/types/crm/lead.ts
asdfryan: will fix in next PR

PR #261, opened 2023-03-13
chore: optimize populateAssociations

albertyfwu:
got rid of the prisma calls for populating associations, paging in records, etc. in favor of a simple single query. runs much much faster and avoids connection pool timeouts, lock contention, etc.

### DIFF
--- a/packages/db/index.ts
+++ b/packages/db/index.ts
@@ -1,8 +1,6 @@
 import { PrismaClient } from '@prisma/client';
 
-const prisma = new PrismaClient({
-  // log: ['query'],
-});
+const prisma = new PrismaClient();
 
 export * from '@prisma/client';
 export default prisma;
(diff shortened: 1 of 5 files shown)


packages/db/index.ts
```diff
@@ -1,6 +1,16 @@
 import { PrismaClient } from '@prisma/client';
 
-const prisma = new PrismaClient();
+const prisma = new PrismaClient({
+  // log: ['query'],
```
asdfryan, line 4: Remove

packages/db/index.ts
```diff
@@ -1,6 +1,16 @@
...
+});
 
 export * from '@prisma/client';
 export default prisma;
+
+// TODO: Shouldn't be hard-coding the DB schema here.
+export const COMMON_MODEL_DB_TABLES = {
```
asdfryan, line 11: There's a user table now

albertyfwu: we can add it when we need it

albertyfwu: ok i'll add it since i have to address the other comment anyway

asdfryan (approved):

PR #264, opened 2023-03-13
fix: add api key API and middleware guards

tomkit:
- add endpoints to generate and revoke an api key
- add middleware to guard crm endpoints with it
- also fix some incorrect openapi examples
- guard all but top-level `/applications/:application_id` endpoints

NOTE:
- we need to add a username/password guard for mgmt endpoints for API or Management Portal access


apps/api/routes/mgmt/v1/application/index.ts
```diff
@@ -87,6 +92,37 @@ export default function init(app: Router): void {
     }
   );
 
+  applicationRouter.post('/:application_id/_generate_api_key', async (req: Request, res: Response) => {
```
albertyfwu: can we use the actual req and res types?

packages/core/types/application.ts
```diff
@@ -8,7 +8,8 @@ export type Application = BaseApplication & {
 };
 
 export type ApplicationConfig = {
-  webhook: WebhookConfig | null;
+  webhook?: WebhookConfig | null;
```
albertyfwu, line 11: what is the reasoning for both `undefined` and `null`? same question for `apiKey`

apps/api/routes/mgmt/v1/application/index.ts
```diff
@@ -87,9 +99,53 @@ export default function init(app: Router): void {
...
+  applicationRouter.post(
+    '/:application_id/_generate_api_key',
+    async (
+      req: Request<CreateApplicationApiKeyPathParams, CreateApplicationApiKeyResponse, CreateApplicationApiKeyRequest>,
+      res: Response<CreateApplicationApiKeyResponse>
+    ) => {
+      const application = await applicationService.getById(req.params.application_id);
+
+      const apiKey = crypto.randomBytes(64).toString('base64');
+      const hashedApiKey = await crypto.scryptSync(apiKey, SUPAGLUE_API_KEY_SALT!, 64).toString('hex'); // TODO: remove bang by getting NodeJs ProcessEnv global interface working
```
asdfryan, line 111: We should move encryption stuff into a common lib (shared with our credentials encryption)

asdfryan: Can move to `lib/crypt.ts`

tomkit: Let's do as another PR? We should also consolidate the secret keys

openapi/mgmt/components/schemas/objects/application_api_key.yaml
```diff
@@ -0,0 +1,5 @@
+type: object
```
asdfryan, line 1: We should add the required api key header too

lucasmarshall (changes requested):

.env.sample
```diff
@@ -1,7 +1,6 @@
-SUPAGLUE_API_KEY_SALT=please-change-me
-
 SUPAGLUE_SYNC_PERIOD_MS=900000
 SUPAGLUE_API_ENCRYPTION_SECRET=CHANGETHIS
+SUPAGLUE_QUICKSTART_API_KEY="8o2kGDTHwTY6zcQV8gx7ep8vV7F2JNdeu95OR75MY7dlQJhzOiM4Ha/dKwXOTfDzDxVo+a6tZBrVxYfzJFIrgw=="
```
asdfryan, line 3: I think the quotes are unnecessary

docker-compose.yml
```diff
@@ -33,6 +33,7 @@ services:
       - SUPAGLUE_DISABLE_ERROR_REPORTING
       - SUPAGLUE_POSTHOG_API_KEY
       - SUPAGLUE_API_ENCRYPTION_SECRET
+      - SUPAGLUE_API_KEY_SALT
```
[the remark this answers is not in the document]
tomkit: I consolidated into SUPAGLUE_API_ENCRYPTION_SECRET

asdfryan (approved):

docker-compose.yml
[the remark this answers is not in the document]
lucasmarshall: Not sure what you mean?

tomkit:
landing, then we can make additional changes later

PR #270, opened 2023-03-14
feat: custom field writes

asdfryan:


openapi/crm/components/schemas/objects/custom_fields.yaml
```diff
@@ -0,0 +1,3 @@
+type: object
+additionalProperties: true
+description: Custom properties to be inserted that is not covered by the common model. Object keys must match exactly to the corresponding provider API.
```
albertyfwu, line 3: nit: that _are_ not

albertyfwu (approved):

albertyfwu (changes requested):
i see a bunch of new `custom_fields` properties everywhere. however, where is the code that actually writes these new fields to hubspot?

albertyfwu (approved):

albertyfwu (approved):

PR #273, opened 2023-03-14
chore: internal routes, unroll management routes

asdfryan:


apps/api/middleware/internal.ts
```diff
@@ -0,0 +1,17 @@
+import { UnauthorizedError } from '@supaglue/core/errors';
+import { NextFunction, Request, Response } from 'express';
+import { getDependencyContainer } from '../dependency_container';
+
+const { applicationService } = getDependencyContainer();
+
+export async function internalMiddleware(req: Request, res: Response, next: NextFunction) {
```
asdfryan, line 7: This will be properly implemented in next PR. I just copied the api key middleware for now

apps/api/middleware/internal.ts
```diff
@@ -0,0 +1,17 @@
...
+  const apiKey = req.headers['x-api-key'] as string;
```
tomkit, line 8: can we call it something other than the public one, e.g. `x-sg-api-key` or something

apps/api/routes/internal/index.ts
```diff
@@ -0,0 +1,10 @@
+import { Router } from 'express';
+import v1 from './v1';
+
+export default function init(app: Router): void {
+  const internalRouter = Router();
+
+  v1(internalRouter);
+
+  app.use('/mgmt', internalRouter);
```
tomkit, line 9: `internal_mgmt` or something like that?

apps/api/middleware/internal.ts
asdfryan: This will be rewritten

apps/api/routes/internal/index.ts
asdfryan: Oops this should be `internal`

tomkit (approved):

apps/api/middleware/internal.ts
asdfryan: I went ahead and implemented the shared token

tomkit (approved):

PR #275, opened 2023-03-14
fix: implement initial version of next-auth

tomkit:
- integrate next-auth (username credentials only)
- add sg_user prisma table and model
- add internal _login endpoint

TODO: 
- update mutates to go through API routes too


apps/api/routes/internal/v1/index.ts
```diff
@@ -1,16 +1,19 @@
...
 import { Router } from 'express';
-import api_key from './api_key';
+import apiKey from './api_key';
+import v1Auth from './auth';
 import customer from './customer';
 import integration from './integration';
 import webhook from './webhook';
 
 export default function init(app: Router): void {
+  v1Auth(app);
```
asdfryan, line 10: This should also be guarded by the internal middleware

apps/mgmt-ui/src/components/AccountMenu/index.tsx
```diff
@@ -71,7 +70,7 @@ export default function AccountMenu() {
         transformOrigin={{ horizontal: 'right', vertical: 'top' }}
         anchorOrigin={{ horizontal: 'right', vertical: 'bottom' }}
       >
-        <MenuItem onClick={handleClose}>
+        {/* <MenuItem onClick={handleClose}>
```
asdfryan, line 73: Why not just delete?

apps/mgmt-ui/src/hooks/useCustomers.ts
```diff
@@ -1,13 +1,12 @@
 import { API_HOST, APPLICATION_ID } from '@/client';
-import { camelcaseKeys } from '@/utils/camelcase';
 import useSWR from 'swr';
 import { fetcher } from '.';
 
 export function useCustomers() {
   const { data, error, isLoading } = useSWR(`${API_HOST}/mgmt/v1/applications/${APPLICATION_ID}/customers`, fetcher);
```
asdfryan, line 6: This should be using internal

packages/db/prisma/schema.prisma
```diff
@@ -18,6 +32,7 @@ model Application {
   updatedAt   DateTime      @updatedAt @map("updated_at")
   Customer    Customer[]
   Integration Integration[]
+  SgUser      SgUser[]
```
asdfryan, line 35: This doesn't feel right -- the way this relation is defined, a single Application can have multiple SgUsers?

apps/mgmt-ui/package.json
```diff
@@ -15,6 +15,7 @@
     "@mui/material": "^5.11.12",
     "@mui/x-data-grid": "^6.0.0",
     "@mui/x-date-pickers": "^6.0.0",
+    "@next-auth/prisma-adapter": "^1.0.5",
```
asdfryan, line 18: Do we need this?

packages/db/prisma/schema.prisma
tomkit: I think we eventually want many-many. For now, I think we’ll have more users invited to the same application since we don’t allow creation of Applications

apps/api/routes/internal/v1/index.ts
tomkit: +1

apps/mgmt-ui/src/hooks/useCustomers.ts
tomkit: +1

apps/mgmt-ui/src/components/AccountMenu/index.tsx
tomkit: +1

packages/db/prisma/schema.prisma
tomkit: I think we eventually want many-many. For now, I think we’ll have more users invited to the same application since we don’t allow creation of Applications

apps/mgmt-ui/package.json
tomkit: Removed

apps/mgmt-ui/src/components/AccountMenu/index.tsx
tomkit: +1

apps/mgmt-ui/.env.sample
```diff
@@ -0,0 +1,3 @@
+SUPAGLUE_JWT_SECRET=a-secret-to-change
+NEXT_PUBLIC_SUPAGLUE_INTERNAL_TOKEN=
```
lucasmarshall, line 2: We can't have this in the FE or we'll have to build a docker image per customer with their shared secret baked in.

lucasmarshall (changes requested):

apps/mgmt-ui/.env.sample
lucasmarshall: We need all api access proxied through the edge so we can inject this.

apps/mgmt-ui/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,46 @@
+import { API_HOST, APPLICATION_ID } from '@/client';
+import NextAuth from 'next-auth';
+import CredentialsProvider from 'next-auth/providers/credentials';
+
+const sgInternalToken = process.env.SUPAGLUE_INTERNAL_TOKEN!;
```
Comment (lucasmarshall) [inline apps/mgmt-ui/src/pages/api/auth/[...nextauth].ts:5]: This isn't being set in the next app

apps/mgmt-ui/.env.sample
tomkit: I fixed this

apps/mgmt-ui/src/pages/api/auth/[...nextauth].ts
Comment (tomkit) [reply apps/mgmt-ui/src/pages/api/auth/[...nextauth].ts:5]: Fixed in the last commit

lucasmarshall (approved):

PR #277, opened 2023-03-15
feat: equality filter on contact email and account website

albertyfwu:


packages/core/services/common_models/contact_service.ts
```diff
@@ -38,6 +40,35 @@ export class ContactService extends CommonModelBaseService {
...
+    const { page_size, cursor } = paginationParams;
+    const pageSize = page_size ? parseInt(page_size) : undefined;
+    const models = await this.prisma.crmContact.findMany({
+      ...getPaginationParams(pageSize, cursor),
+      where: {
+        connectionId,
+        emailAddresses:
+          filters.emailAddress?.type === 'equals'
+            ? {
+                array_contains: [{ emailAddress: filters.emailAddress.value }],
```
albertyfwu, line 57: ideally we just pass in `path` directly to `$[*].emailAddress == '${filters.emailAddress.value}'` but prisma doesn't support that, so we need to do `array_contains`. It looks like it's working.

packages/core/types/filter.ts
```diff
@@ -0,0 +1,12 @@
+export type EqualsFilter = {
+  type: 'equals';
+  value: string;
+};
+
+// export type ContainsFilter = {
```
albertyfwu, line 6: 'equals' was already stretching the capabilities of prisma so i didn't implement contains.

didn't want to implement filtering + pagination together in raw query. down the road we may want to reconsider using prisma if we are always sidestepping it for common model operations

tomkit (approved):

PR #285, opened 2023-03-15
fix: fix open api types

asdfryan:


openapi/crm/components/schemas/objects/addresses.yaml
```diff
@@ -26,19 +26,27 @@ items:
       type: string
       nullable: true
       example: CA
-    street_1:
+    street1:
```
asdfryan, line 29: the snakecase of street1 is just street1, not street_1

tomkit (approved):

PR #287, opened 2023-03-16
fix: add api key to mgmt ui

tomkit:
- add api key tab to configuration screen
- add swr and api route for api key
- add a path to url to differentiate between /configure/integrations and /configure/api_keys
- fix seed casing for api key
- return raw api key upon creation

Screenshot 2023-03-16 at 12 24 54 AM


asdfryan (dismissed):

tomkit:
Landing based on previous approval (fixed rebase)

PR #296, opened 2023-03-16
chore: helm chart and fixes

lucasmarshall:


helm/GKE_RUNBOOK.md
```diff
@@ -0,0 +1,129 @@
...
+Set up your environment variables:
+
+```shell
+export DEPLOYMENT_ID=<a unique id for this deployment, usually your company name>
+export PROJECT_ID=<your GCS project id>
+export REGION=<your region>
+export CLUSTER_NAME=<your GKE cluster>
+export API_HOST=supaglue-api.<your domain>
+export MANAGEMENT_HOST=supaglue-management.<your domain>
+export OAUTH_RETURN_URL=<the url you want users to return to by default after auth (usually your app)>
```
tomkit, line 23: i guess this is actually a fallback/default since it can be specified in the embedded link too

packages/db/prisma/seed.ts
```diff
@@ -161,27 +161,36 @@ export async function cryptoHash(text: string): Promise<{ original: string; hash
 }
 
 async function seedApplication() {
-  const { hashed: hashedApiKey } = await cryptoHash(SUPAGLUE_QUICKSTART_API_KEY!);
+  let webhook = {};
 
+  if (process.env.NODE_ENV === 'development') {
+    webhook = {
```
asdfryan, line 167: IMO We should just remove the webhook seed (set it to null) for all cases.

lucasmarshall: Will do

packages/db/prisma/seed.ts
```diff
@@ -182,15 +170,15 @@ async function seedApplication() {
       name: 'My App',
       config: {
         api_key: '',
-        webhook,
+        webhook: {},
```
asdfryan, line 173: According to our types it looks like the `webhook` field should either be `null` or a complete `WebhookConfig` -- so I think this should be null?

tomkit (approved):
rubberstamping

tomkit (approved):

tomkit (approved):
nit: maybe we can have two diff seeds down the road

PR #310, opened 2023-03-17
fix: PATCH for all CRM objects for hubspot

asdfryan:
Our PATCH was more of a PUT -- empty properties were getting nulled out. This PR fixes this.

Tested locally:
Screenshot 2023-03-17 at 10 28 08 AM


packages/core/remotes/crm/hubspot/mappers.ts
```diff
@@ -216,44 +216,50 @@ const getFullName = (firstName?: string, lastName?: string): string | null => {
...
+  const out = {
+    firstname: nullToEmptyString(params.firstName),
+    lastname: nullToEmptyString(params.lastName),
+    hubspot_owner_id: nullToEmptyString(params.ownerId),
     ...toHubspotEmailCreateParams(params.emailAddresses),
     ...toHubspotPhoneCreateParams(params.phoneNumbers),
     ...toHubspotAddressCreateParams(params.addresses),
     ...params.customFields,
   };
+  removeUndefinedValues(out);
```
albertyfwu, line 261: is this necessary?

albertyfwu (commented):
do we need to do this for salesforce too?

packages/core/remotes/crm/hubspot/mappers.ts
asdfryan: yes

albertyfwu (approved):

PR #317, opened 2023-03-17
chore: document webhook payload

lucasmarshall:


openapi/mgmt/components/schemas/webhook-payload.yaml
```diff
@@ -0,0 +1,41 @@
+oneOf:
+  - type: object
+    title: Sync Finished Event
+    properties:
+      type:
```
albertyfwu, line 5: do we want to include `required:` to be more stricter?

albertyfwu (approved):

openapi/mgmt/components/schemas/webhook-payload.yaml
```diff
@@ -0,0 +1,55 @@
...
+        enum:
+          - SYNC_SUCCESS
+          - SYNC_ERROR
+      payload:
+        type: object
+        properties:
+          connectionId:
```
albertyfwu, line 12: i think we are using camelcase as convention for other endpoints

lucasmarshall: It doesn't look like we snake-case when we send the webhook, so this is accurate. Are you suggesting we change the webhook payloads to snake-case?

albertyfwu: yea wondering if we should be consistent, but i don't feel super strongly

albertyfwu (approved):

asdfryan (dismissed):

openapi/mgmt/openapi.yaml
```diff
@@ -61,3 +61,20 @@ components:
       $ref: ./components/schemas/create_update_integration.yaml
     webhook:
       $ref: ./components/schemas/objects/webhook.yaml
+    webhook-payload:
+      $ref: './components/schemas/webhook-payload.yaml'
+webhooks:
```
tomkit, line 66: move into paths?

lucasmarshall: It's not a path

lucasmarshall: We can move it into a new webhooks directory if you like

openapi/crm/openapi.yaml
```diff
@@ -1,4 +1,4 @@
-openapi: 3.0.3
+openapi: 3.1.0
```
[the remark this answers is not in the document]
lucasmarshall: I didn't observe any. I'll do some more testing though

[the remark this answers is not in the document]
albertyfwu: the code has a regex that checks for 3.0.3 and i was running into that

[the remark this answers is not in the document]
lucasmarshall: Yeah, but it still works. I'll do a hack so it doesn't complain though.

albertyfwu (approved):

PR #318, opened 2023-03-17
feat: move sync_info/sync_history to mgmt, support application-wide querying

asdfryan:
* move `sync-info` and `sync-history` into `mgmt/v1`
* refactor openAPI spec
* refactor / rewrite services to require `applicationId` in `list()` methods for mgmt services
* add connectionService as dependency to `syncService` and `syncHistoryService`
* change `SyncInfo` and `SyncHistory` types to contain more contextual info (`customerId`, `connectionId`, `providerName`, `category`)
* various other refactors


lucasmarshall:
We probably also need to update docs?

lucasmarshall (dismissed):

packages/core/services/connection_service.ts
```diff
@@ -3,42 +3,43 @@ import { NotFoundError } from '../errors';
...
   }
 
   public async getSafeById(id: string): Promise<ConnectionSafe> {
-    const integration = await this.#prisma.connection.findUnique({
+    const connection = await this.#prisma.connection.findUnique({
       where: { id },
     });
-    if (!integration) {
+    if (!connection) {
       throw new NotFoundError(`Can't find integration with id: ${id}`);
```
albertyfwu, line 32: ditto

packages/core/services/connection_service.ts
```diff
@@ -3,42 +3,43 @@ import { NotFoundError } from '../errors';
...
-  // TODO: paginate
-  public async listUnsafe(): Promise<ConnectionUnsafe[]> {
-    const connections = await this.#prisma.connection.findMany();
-    return connections.map((connection) => fromConnectionModelToConnectionUnsafe(connection));
+    return fromConnectionModelToConnectionSafe(connection);
   }
 
-  public async listSafe(): Promise<ConnectionSafe[]> {
-    const connections = await this.#prisma.connection.findMany();
+  public async listSafe(applicationId: string, customerId?: string, providerName?: string): Promise<ConnectionSafe[]> {
```
albertyfwu, line 37: when do we need to use this endpoint? Should we provide a more scoped endpoint that doesn't just list all connections in an application?

packages/core/services/integration_service.ts
```diff
@@ -33,8 +33,8 @@ export class IntegrationService {
   }
 
   // TODO: paginate
-  public async list(): Promise<Integration[]> {
-    const integrations = await this.#prisma.integration.findMany();
+  public async list(applicationId: string): Promise<Integration[]> {
+    const integrations = await this.#prisma.integration.findMany({ where: { applicationId } });
     return integrations.map((integration) => fromIntegrationModel(integration));
```
albertyfwu, line 38: nit: `return integrations.map(fromIntegrationModel)`

apps/api/routes/internal/v1/sync_history.ts
```diff
@@ -18,9 +18,11 @@ export default function init(app: Router) {
       res: Response<GetSyncHistoryResponse>
     ) => {
       const { next, previous, results } = await syncHistoryService.list({
-        connectionId: req.customerConnection.id,
-        paginationParams: req.query,
-        model: req.query.model,
+        applicationId: req.supaglueApplication.id,
```
albertyfwu, line 21: why do we need sync history and sync info to move under `mgmt`? the advantage of keeping it under `crm` is that the developer can pass in the same headers they already pass in for the common model endpoints

packages/core/services/connection_service.ts
asdfryan: We need to use this in `sync_history_service.ts`

apps/api/routes/internal/v1/sync_history.ts
asdfryan: See https://linear.app/supaglue/issue/SUP1-150/add-back-sync-logs-to-management-portal for rationales for why.

We need to be able to query for sync history independent of customerId/providerName

apps/api/services/sync_service.ts
```diff
@@ -7,12 +8,23 @@ import { Client, ScheduleAlreadyRunning } from '@temporalio/client';
...
+  #connectionService: ConnectionService;
 
-  public constructor(temporalClient: Client) {
+  public constructor(temporalClient: Client, connectionService: ConnectionService) {
     this.#temporalClient = temporalClient;
+    this.#connectionService = connectionService;
   }
 
-  public async getSyncInfo(connectionId: string, commonModel: CommonModel): Promise<SyncInfo> {
+  public async getSyncInfoList(applicationId: string, customerId?: string, providerName?: string): Promise<SyncInfo[]> {
```
albertyfwu, line 18: what do you think about nesting `customerId?: string` and `providerName?: string` under something like `filters`? Seems cleaner when you want to just filter by provider and not customer.

apps/api/services/sync_service.ts
```diff
@@ -7,12 +8,23 @@ import { Client, ScheduleAlreadyRunning } from '@temporalio/client';
...
+    const connections = await this.#connectionService.listSafe(applicationId, customerId, providerName);
+    const out = await Promise.all(connections.flatMap((connection) => this.getSyncInfoListFromConnection(connection)));
```
albertyfwu, line 20: not super critical right now, but i wonder if there is some bulk endpoint for temporal here. it looks like we could issue 100s, 1000s, etc. of requests to temporal at the same time.

apps/api/routes/internal/v1/sync_history.ts
```diff
@@ -18,9 +18,11 @@ export default function init(app: Router) {
...
+        paginationParams: { page_size: req.query?.page_size, cursor: req.query?.cursor },
```
albertyfwu, line 22: can `query` be undefined here? why do we need `?.`?

apps/api/routes/mgmt/v1/sync_history.ts
```diff
@@ -0,0 +1,38 @@
...
+export default function init(app: Router) {
+  app.get(
+    '/sync-history',
+    async (
+      req: Request<GetSyncHistoryPathParams, GetSyncHistoryResponse, GetSyncHistoryRequest, GetSyncHistoryQueryParams>,
+      res: Response<GetSyncHistoryResponse>
+    ) => {
+      const { next, previous, results } = await syncHistoryService.list({
+        applicationId: req.supaglueApplication.id,
+        paginationParams: { page_size: req.query?.page_size, cursor: req.query?.cursor },
```
albertyfwu, line 22: same q here about `query?`.

albertyfwu (approved):

apps/api/routes/internal/v1/sync_history.ts
asdfryan: It can indeed be undefined, according to the generated types from our open API

apps/api/services/sync_service.ts
asdfryan: Unfortunately not. There is a `list()` method, but it only allows you to fetch all existing schedules.

PR #322, opened 2023-03-18
chore: a quick pass at mgmt-ui cleanup

tomkit:
- use some types
- remove unused code


apps/mgmt-ui/src/components/configuration/VerticalTabs.tsx
```diff
@@ -113,21 +59,14 @@ function TabPanel(props: TabPanelProps) {
...
 }
 
-function a11yProps(index: number) {
-  return {
-    id: `vertical-tab-${index}`,
-    'aria-controls': `vertical-tabpanel-${index}`,
-  };
-}
-
 export default function VerticalTabs() {
```
asdfryan, line 62: Can you rename as there are no longer vertical tabs?

apps/mgmt-ui/src/components/configuration/VerticalTabs.tsx
```diff
@@ -113,21 +59,14 @@ function TabPanel(props: TabPanelProps) {
...
   const router = useRouter();
   const { tab = [] } = router.query;
```
asdfryan: I found this code difficult to understand. 

I think it will be more readable if you just pass in `providerName` in the props instead

apps/mgmt-ui/src/components/configuration/IntegrationTabPanel.tsx
```diff
@@ -1,31 +1,154 @@
...
   status: string;
 };
 
-export default function IntegrationTabPanel(props: IntegrationTabPanelProps) {
-  const { integrationCardsInfo, existingIntegrations, status } = props;
+export default function IntegrationTabPanel(props: IntegrationDetailTabPanelProps) {
+  const { applicationId } = useApplication();
+  const { integration, integrationCardInfo } = props;
+  const [clientId, setClientId] = useState('');
+  const [clientSecret, setClientSecret] = useState('');
```
asdfryan, line 23: Could you set types? i.e. `useState<string>`

tomkit: TS infers it using the default value, but I can be explicit

apps/mgmt-ui/src/pages/api/internal/integrations/[...integrationId].ts
```diff
@@ -12,7 +13,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse<
   });
 
   if (!result.ok) {
-    return res.status(500).json({ error: 'Failed to fetch' });
+    return res.status(500).json(null);
```
Comment (asdfryan) [inline apps/mgmt-ui/src/pages/api/internal/integrations/[...integrationId].ts:16]: Why are we no longer passing an error?

Comment (tomkit) [reply apps/mgmt-ui/src/pages/api/internal/integrations/[...integrationId].ts:16]: It was an arbitrary response I chose when implementing last week. In terms of request/response signatures there's more to do so just punting in this pass of cleanup

asdfryan (approved):

PR #325, opened 2023-03-20
chore: remove relationship between SgUser and Application

albertyfwu:
For non-Cloud, any `SgUser` can access any `Application` and there are no roles. This gets rid of the relationship (to avoid dealing with roles, M-M mapping between SGUser and Application, etc.)


tomkit (approved):

tomkit (dismissed):
I think there are other places where we are making actions against the default seeded Application

tomkit (approved):

PR #327, opened 2023-03-20
feat: update openAPI spec, middleware, and endpoints for customer_id (external)

asdfryan:


apps/api/middleware/connection.ts
```diff
@@ -2,19 +2,21 @@ import { UnauthorizedError } from '@supaglue/core/errors';
...
     throw new UnauthorizedError(`x-customer-id and x-provider-name headers must be set`);
   }
 
+  const customer = await customerService.getByExternalId(req.supaglueApplication.id, externalCustomerId);
+
   const integration = await integrationService.getByProviderName(providerName);
 
   req.customerConnection = await connectionService.getSafeByCustomerIdAndIntegrationId({
-    customerId,
+    customerId: customer.id,
```
tomkit, line 19: is `getSafeByCustomerIdAndIntegrationId` already using customerId as `externalIdentifier` ?

asdfryan: No. The service still expects the internal id (as that is what the unique constraint is set on)

PR #333, opened 2023-03-20
feat: applications routes

albertyfwu:
for creating, listing, getting


asdfryan (commented):
I think you'll also need to dup the internal endpoints as well

apps/api/routes/internal/v1/index.ts
```diff
@@ -14,6 +15,7 @@ export default function init(app: Router): void {
 
   auth(v1Router);
   apiKey(v1Router);
+  application(v1Router);
```
asdfryan, line 18: This is awkward because this uses the same `internalMiddlware` which identifies the `supaglueApplication`.

I think you should probably either create a new middleware that does the shared token authentication but does not identify the `supaglueApplication` or restructure all the routes to be nested under `application`

asdfryan (approved):

PR #336, opened 2023-03-20
feat: customerId as concat of `applicationId:externalId`, misc fixes

asdfryan:
* `Customer.Id` is now `applicationId:externalId` instead of UUID
* `Connection`, `SyncInfo`, `SyncHistory` now contain `applicationId` and `customerId` (the external one)
* misc fixes to open API endpoints

### DIFF
--- a/packages/core/lib/customerid.ts
+++ b/packages/core/lib/customerid.ts
@@ -1,0 +1,16 @@
+import { InternalServerError } from '@supaglue/core/errors';
+
+export const getCustomerId = (applicationId: string, externalCustomerId: string): string => {
+  return `${applicationId}:${externalCustomerId}`;
+};
+
+export const parseCustomerId = (customerId: string): { applicationId: string; externalCustomerId: string } => {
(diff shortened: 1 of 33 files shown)


lucasmarshall:
Are we planning to provide a migration script for customers to run here?

packages/core/lib/customerid.ts
```diff
@@ -0,0 +1,16 @@
+import { InternalServerError } from '@supaglue/core/errors';
```
tomkit, line 1: nit: `customer_id.ts`

tomkit:
nit: `getCustomerId`->`getCustomerIdPk` since it's the hash/union PK, to differentiate between `customerId` which is the external

tomkit (approved):

asdfryan:
> Are we planning to provide a migration script for customers to run here?

Yeah I think we have to. I think @tomkit is working on this.

lucasmarshall (approved):

tomkit (approved):

PR #339, opened 2023-03-21
feat: switch between applications in FE

albertyfwu:
image


apps/mgmt-ui/src/hooks/index.ts
```diff
@@ -1,10 +1,17 @@
-export const APPLICATION_ID = 'a4398523-03a2-42dd-9681-c91e3e2efaf4';
-
 export const fetcher = <T>(input: RequestInfo, init: RequestInit): Promise<T> => {
+  return fetch(input, init).then<T>((res) => res.json());
+};
```
tomkit, line 3: what's the diff between context/fetcher and hooks/fetcher?

apps/mgmt-ui/src/pages/index.tsx
```diff
@@ -21,8 +22,29 @@ export const getServerSideProps: GetServerSideProps = async ({ req, res }) => {
     };
   }
 
+  // Get applications to set active application
```
tomkit, line 25: is this the same call as `apps/mgmt-ui/src/pages/api/internal/applications/index.ts:9` ?

apps/mgmt-ui/src/components/ApplicationMenu/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+          textAlign: 'start',
+          px: 3,
+          color: '#fff',
+          borderRadius: 0,
+        }}
+        onClick={handleClick}
+        endIcon={<KeyboardArrowDownIcon />}
+      >
+        <Box>
+          <div style={{ fontSize: 12 }}>Application</div>
```
tomkit, line 55: the css is a mess right now, but it's between sx and tailwindcss, can you convert inline style prop to one of those

apps/mgmt-ui/src/pages/index.tsx
tomkit: or leave comment its' the same as file ^

tomkit (approved):

apps/mgmt-ui/src/hooks/index.ts
albertyfwu: no difference. i just didn't want to import from the other directory. later, we may want to do a pass at cleaning up our SWR/fetcher uses

PR #355, opened 2023-03-21
feat: unique application name

albertyfwu:


packages/db/prisma/schema.prisma
```diff
@@ -23,7 +23,7 @@ model SgUser {
 
 model Application {
   id          String        @id @default(uuid())
-  name        String
+  name        String        @unique
```
asdfryan, line 26: We need to think more about this IMO.

In a cloud, multi-tenant offering, the name should be unique within an organization only.

albertyfwu: Yes, but we decided to ignore multi-tenant, roles, etc. and just make this work for single-tenant for now.

asdfryan (approved):

Issue #360, opened 2023-03-21
[SUP1-22] Add search attributes for Temporal workflows/activities


albertyfwu:
For searchability/debuggability

<sub>From [SyncLinear.com](https://synclinear.com) | [SUP1-22](https://linear.app/supaglue/issue/SUP1-22/add-search-attributes-for-temporal-workflowsactivities)</sub>

PR #394, opened 2023-03-22
chore: search attributes for temporal workflows

albertyfwu:
https://github.com/supaglue-labs/supaglue/issues/360

We'll also need to make a change to the Helm chart to run `init_temporal.js` as well: https://github.com/supaglue-labs/supaglue/pull/411


docker-compose.override.dev.yml
```diff
@@ -39,4 +39,4 @@ services:
     volumes:
       - .:/app
     working_dir: /app
-    command: /bin/sh -c "yarn workspace @supaglue/db prisma migrate dev && yarn workspace @supaglue/db prisma db seed"
+    command: /bin/sh -c "yarn workspace @supaglue/db prisma migrate dev && yarn workspace @supaglue/db prisma db seed && yarn workspace api init-temporal"
```
albertyfwu, line 42: @lucasmarshall any better way of doing this? for self-hosted customers using Helm chart, they would need to run some setup script

tomkit (approved):

PR #399, opened 2023-03-22
fix: application create

asdfryan:


apps/mgmt-ui/src/pages/api/internal/applications/index.ts
```diff
@@ -33,6 +33,7 @@ export default async function handler(
         },
         body: JSON.stringify({
           name: req.body.name,
+          orgId: ORGANIZATION_ID,
```
albertyfwu, line 36: did you test this? doesn't this need to be `org_id`?

asdfryan: I think it doesn't actually matter because there is no open API validation, and camelcase of `orgId` is still `orgId`.

albertyfwu: I SEE

albertyfwu (approved):

PR #400, opened 2023-03-22
feat: Clerk WIP (draft)

asdfryan:


apps/mgmt-ui/src/components/Logout/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const CloudLogout = () => {
+  const { signOut: clerkSignOut } = useClerk();
+  const handleLogout = async () => {
+    await clerkSignOut();
+  };
+  return <LogoutImpl handleLogout={handleLogout} />;
+};
+
+export const Logout = () => {
```
tomkit, line 33: nit: this could be a ternary

apps/mgmt-ui/src/hooks/useActiveApplication.ts
```diff
@@ -0,0 +1,20 @@
...
+export function useActiveApplication() {
+  const applicationId = useActiveApplicationId();
+  const { data, error, isLoading, ...rest } = useSWR(
+    `/api/internal/applications/${applicationId}`,
+    fetcher<Application>
+  );
+
+  return {
+    activeApplication: data ? camelcaseKeys(data) : undefined,
+    isLoading,
```
albertyfwu, line 16: is it possible to have this be non-nullable and make sure only to call this from components in the tree that MUST be scoped under application? It seems annoying to have all components check if active application exists, even if you know that must be the case, since it's something like `/applications/{applicationId}/customers`

apps/mgmt-ui/src/middleware.ts
```diff
@@ -0,0 +1,55 @@
...
+  const { userId, orgId } = getAuth(request);
+
+  if (!userId) {
+    // redirect the users to /pages/sign-in/[[...index]].ts
+
+    const signInUrl = new URL('/sign-in', request.url);
+    signInUrl.searchParams.set('redirect_url', request.url);
+    return NextResponse.redirect(signInUrl);
+  }
+  if (!orgId && !isCloudCreateOrgPath(request.nextUrl.pathname)) {
```
tomkit, line 45: nit: assign conditional to a variable for readability

apps/mgmt-ui/src/components/configuration/IntegrationTabPanel.tsx
```diff
@@ -151,7 +151,7 @@ export default function IntegrationTabPanel(props: IntegrationDetailTabPanelProp
                   ei.id === newIntegration.id ? newIntegration : ei
                 );
 
-                updateRemoteIntegration(activeApplication.id, newIntegration);
+                updateRemoteIntegration(activeApplicationId as string, newIntegration);
```
albertyfwu, line 154: is there a way to avoid doing this everywhere?

asdfryan: Not sure as this is at the component level, and there is no guarantee that the router has `applicationId` set. We can throw an exception I guess, but is that really what we want?

asdfryan: I'm throwing an exception

apps/mgmt-ui/src/hooks/useActiveApplication.ts
asdfryan: Well the problem is this relies on a server fetch which might fail, so it has to be undefined.

The proper way to deal with this is to condition on `isLoading` or `isError` everywhere.

apps/mgmt-ui/src/middleware.ts
asdfryan: I added comments instead

tomkit (approved):

PR #404, opened 2023-03-23
fix: disable error reporting in management UI in dev env

lucasmarshall:


albertyfwu (approved):
not directly related to this PR, but i noticed before that if you set something like `SUPAGLUE_DISABLE_ERROR_REPORTING` to `0`, it will get read in as a string and not be falsey. not sure how obvious this is to users

PR #408, opened 2023-03-23
fix: customerId should be webhook payload

asdfryan:


packages/core/types/application.ts
```diff
@@ -30,3 +30,9 @@ export type ApplicationCreateParams = {
   orgId: string;
 };
 export type ApplicationUpdateParams = Partial<BaseApplication>;
+
+export type WebhookPayloadType = 'CONNECTION_SUCCESS' | 'CONNECTION_ERROR' | 'SYNC_SUCCESS' | 'SYNC_ERROR';
+
+export type WebhookPayload = {
+  customerId: string;
+} & Record<string, any>;
```
albertyfwu, line 38: can we actually type this? this is an important part of our API

albertyfwu (approved):

Issue #407, opened 2023-03-23
[SUP1-207] Migration docs to v0.6.0


tomkit:
Docs5

<sub>From [SyncLinear.com](https://synclinear.com) | [SUP1-207](https://linear.app/supaglue/issue/SUP1-207/migration-docs-to-v060)</sub>

PR #409, opened 2023-03-23
docs: upgrade instructions for 0.6.0

albertyfwu:
https://github.com/supaglue-labs/supaglue/issues/407

### DIFF
--- a/helm/GKE_RUNBOOK.md
+++ b/helm/GKE_RUNBOOK.md
@@ -125,31 +125,33 @@ Go to
 
 ## Upgrading
 
-### Upgrading to 0.5.x
+### Upgrading to 0.6.x
 
-We have some backward-incompatible changes in 0.5.x. You will need to run the following process to upgrade.
+We have some backward-incompatible changes in 0.6.x. You will need to run the following process to upgrade.
 
 ```shell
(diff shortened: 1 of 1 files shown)


helm/GKE_RUNBOOK.md
```diff
@@ -125,31 +125,33 @@ Go to
...
-kubectl port-forward statefulset/supaglue-postgresql 5432:5432 -n supaglue
+kubectl port-forward svc/supaglue-postgresql 5432:5432 -n supaglue
 ```
 
 Then in a new terminal:
 
 ```shell
 export PGPASSWORD=$(kubectl get secrets supaglue-postgresql -n supaglue -o jsonpath='{.data.password}' | base64 --decode)
-psql -U supaglue -d supaglue -h localhost -c "TRUNCATE customers CASCADE;"
+psql -U supaglue -d supaglue -h localhost -c "DROP SCHEMA public CASCADE;"
```
asdfryan, line 151: Isn't our schema `api` not `public`?

albertyfwu: locally it's `api`, but i saw on our k8s cluster it's `public`. @lucasmarshall where do we set the db schema for helm deploys?

albertyfwu: in helm template it doesnt set schema, so by default it is `public`

PR #415, opened 2023-03-23
fix: refresh auth on passthrough

albertyfwu:


packages/core/remotes/base.ts
```diff
@@ -29,9 +29,15 @@ export abstract class AbstractRemoteClient extends EventEmitter implements Remot
...
   }
 
+  protected async maybeRefreshAuth(): Promise<void> {
+    return;
+  }
+
   protected abstract getAuthHeadersForPassthroughRequest(): Record<string, string>;
 
   public async sendPassthroughRequest(request: SendPassthroughRequestRequest): Promise<SendPassthroughRequestResponse> {
+    await this.maybeRefreshAuth();
```
asdfryan, line 39: I think we should override the HubspotClient instead

albertyfwu: fixed

asdfryan (approved):

PR #416, opened 2023-03-23
chore: v0.6.0

lucasmarshall:


albertyfwu (commented):
Could you add `&& node init_scripts/init_temporal.js` to `helm/templates/schema-job.yaml`?

image

PR #442, opened 2023-03-23
feat: org-scoped internal application endpoints

asdfryan:
* `x-org-id` as header for list/create application internal requests
* Don't seed applications for cloud
* create application lazily if one doesn't exist for the org


tomkit:
Is there a place we expose org id so developers know what id to use?

asdfryan:
> Is there a place we expose org id so developers know what id to use?

Discussed offline. It's not necessary.

albertyfwu (approved):

PR #458, opened 2023-03-27
feat: remote_account_id on connection

asdfryan:
All `connection` objects now have a `remote_account_id` representing the account Id in the remote provider (see screenshot below)

Screenshot 2023-03-27 at 2 56 45 PM

### Migration instructions
Please reset all your connections, or hit the `{{baseUrl}}/mgmt/v1/_backfill_remote_account_ids` endpoint to backfill the `remote_account_id`s


albertyfwu (commented):
looks good, but can we pick a name that's less confusing? we also have `remoteAccountId` on crm contacts

albertyfwu (approved):

PR #460, opened 2023-03-27
feat: create integrations in UI

asdfryan:
Also:
- added loading states in various related pages
- changed redirects to already include applicationId scoping for faster redirects

https://www.loom.com/share/a026a7eec0d4465f885a7270fcf146f0

### DIFF
--- a/apps/mgmt-ui/src/client.ts
+++ b/apps/mgmt-ui/src/client.ts
@@ -1,5 +1,5 @@
 import { snakecaseKeys, snakecaseKeysSansHeaders } from '@supaglue/core/lib/snakecase';
-import { Application, Integration, WebhookConfig } from '@supaglue/core/types';
+import { Application, CRMIntegrationCreateParams, Integration, WebhookConfig } from '@supaglue/core/types';
 
 // TODO: use Supaglue TS client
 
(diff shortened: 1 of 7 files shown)


apps/mgmt-ui/src/client.ts
```diff
@@ -53,6 +53,23 @@ export async function deleteWebhook(applicationId: string): Promise<void> {
...
 
+export async function createRemoteIntegration(
+  applicationId: string,
+  data: CRMIntegrationCreateParams
+): Promise<Integration> {
+  const result = await fetch(`/api/internal/integrations/create`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      'x-application-id': applicationId,
```
albertyfwu, line 64: who checks that the user is allowed to create integrations for this `applicationId`? This is in the FE code and not in the next.js lambdas. Can't anybody hit this endpoint, pass in an applicationId, and do this even for orgs they don't own?

apps/mgmt-ui/src/components/configuration/IntegrationCard.tsx
```diff
@@ -40,7 +53,7 @@ export default function IntegrationCard(props: { integration: Integration; integ
           <Button
             variant="text"
             onClick={() => {
-              router.push(`/configuration/integrations/${category}/${providerName}`);
+              router.push(`/applications/${applicationId}/configuration/integrations/${category}/${providerName}`);
```
albertyfwu, line 56: in the future maybe we can create another hook that automatically prepends this for simplicity

apps/mgmt-ui/src/client.ts
asdfryan: There will be a follow-up PR

albertyfwu: https://github.com/supaglue-labs/supaglue/issues/465

albertyfwu (approved):

PR #461, opened 2023-03-28
feat!: [Part 1] introduce Sync object/strategy/phase and implement full refresh

albertyfwu:
**This is a breaking change. You will need to re-connect customer connections to kick off new syncs. We could provide a migration script to automatically backfill `Sync` DB objects if you have customer connections already that you don't want to re-connect.**

This introduces the type definition for the new `Sync` object.

It also implements full refresh sync.

Next PR will implement incremental sync.


asdfryan (commented):
Looks like some builds are failing?

packages/core/types/sync.ts
```diff
@@ -0,0 +1,80 @@
...
+ *       reverse chron order until March 1. We will then ensure that all associations are
+ *       created, and notify the user.
+ *    c: sync records starting from March 1 (for all common models) in
+ *       reverse chron order until the beginning of time. We will then ensure that all
+ *       associations are created, and notify the user.
+ * 2. On subsequent syncs, we will sync records starting from March 28 (for all common models)
+ *    in forward chron order until the end of time. We will then ensure that all associations
+ *    are created, and notify the user. We keep track of the high-water mark and keep doing this.
+ */
+export type ReverseThenForwardSync = BaseSync & {
```
albertyfwu, line 40: i included this even though we're not implementing it because we only decided today that we're not going to to 'reverse then forward' and only 'full then incremental'

apps/sync-worker/services/sync_service.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export class SyncService {
+  #prisma: PrismaClient;
+
+  public constructor(prisma: PrismaClient) {
+    this.#prisma = prisma;
+  }
+
+  public async getSyncById(id: string): Promise<Sync> {
```
tomkit, line 26: is it intentional to have the same methods in sync-worker and api?

packages/core/types/sync.ts
```diff
@@ -0,0 +1,80 @@
+import { CommonModel } from './common';
+
+type BaseSync = {
+  id: string;
+  connectionId: string;
+};
+
+export type FullThenIncrementalSync = BaseSync & {
+  type: 'full then incremental';
```
tomkit, line 9: nit: snakecase in case we need to use this as a constant/enum somewhere that doesn't allow spaces?

packages/core/types/sync.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+export type ReverseThenForwardSync = BaseSync & {
+  type: 'reverse then forward';
+  // These timestamps should be provided in reverse chronological order, e.g. [March 15, March 1, February 23].
+  reversePhaseCutoffTimestampMsList: number[];
+  state: ReverseThenForwardSyncState;
+};
+
+export type SyncType = 'full then incremental' | 'reverse then forward';
+export type Sync = FullThenIncrementalSync | ReverseThenForwardSync;
```
tomkit, line 48: Rather than encode the strategy description into the type name, thoughts on calling them Strategy1, Strategy2, and then having a comment code-block describing them?

I'm having a bit of trouble keeping track of forward/reverse and their permutations

packages/core/types/sync.ts
```diff
@@ -0,0 +1,80 @@
...
+  cutoffTimestampMs: number;
+  status: 'in progress' | 'done';
+};
+type ReverseThenForwardSyncStateForwardPhase = {
+  phase: 'forward';
+  status: 'in progress' | 'done';
+  maxSyncedObjectTimestampMsMap: Record<CommonModel, number>;
+};
+export type ReverseThenForwardSyncState =
+  | ReverseThenForwardSyncStateCreatedPhase
```
tomkit, line 76: Similar to the comment above, thoughts on calling it something like Strategy1Phase1, Strategy1Phase2, and then describe it in comments?

packages/sync-workflows/workflows/run_syncs.ts
```diff
@@ -1,62 +1,254 @@
...
+    switch (sync.type) {
+      case 'full then incremental':
+        numRecordsSyncedMap = await doFullThenIncrementalSync({ sync });
+        break;
+      case 'reverse then forward':
+        numRecordsSyncedMap = await doReverseThenForwardSync({ sync });
+        break;
+    }
+  } catch (err: any) {
+    await Promise.all(
```
tomkit, line 54: I guess another part of the strategy is whether to try to complete k objects as fast as possible first or to make slower progress across all objects we support?

apps/sync-worker/services/sync_service.ts
albertyfwu: yea this single method was intentional, though later we should share it. i didn't want the `api` to have access to updating the sync state. only the `sync-worker` should be allowed to do that

packages/core/types/sync.ts
albertyfwu: where would we need to use this that doesn't allow spaces?

packages/core/types/sync.ts
albertyfwu: is that really clearer? elsewhere in the code we won't know what strategy3, phase 4 mean

packages/sync-workflows/workflows/run_syncs.ts
albertyfwu: yea we could support that later

packages/core/types/sync.ts
tomkit: Might be a personal preference; not a strong point of feedback anyhow

tomkit (approved):

Issue #306, opened 2023-03-17
[SUP1-90] Incremental Hubspot syncs


tomkit:
<sub>From [SyncLinear.com](https://synclinear.com) | [SUP1-90](https://linear.app/supaglue/issue/SUP1-90/incremental-hubspot-syncs)</sub>

tomkit:
We'll see how the current refresh latency works for them. I think for \~30K records, we'll be OK, but if they have closer to what we tested (\~230K) then the latency becomes 10 mins rather than 5 mins

PR #478, opened 2023-03-28
feat: [Part 2] incremental syncs in hubspot

albertyfwu:
https://github.com/supaglue-labs/supaglue/issues/306

1. For all common models in hubspot, this first calls the search endpoint to get new records.
2. For deals and contacts in hubspot, this also calls the association endpoint to get all associated companies for those two object types.

### DIFF
--- a/packages/core/remotes/crm/hubspot/index.ts
+++ b/packages/core/remotes/crm/hubspot/index.ts
@@ -79,7 +79,7 @@ const propertiesToFetch = {
     'hs_createdate', // TODO: Use this or createdate?
     'hs_is_contact', // TODO: distinguish from "visitor"?
     'hubspot_owner_id',
-    'lastmodifieddate',
+    'lastmodifieddate', // hs_lastmodifieddate is missing
     'lastname',
     'mobilephone',
     'phone',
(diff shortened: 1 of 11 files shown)


packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -352,6 +455,65 @@ class HubSpotClient extends AbstractCrmRemoteClient {
...
+      try {
+        await this.maybeRefreshAccessToken();
+
+        // Get contacts
+        const response = await this.#client.crm.contacts.searchApi.doSearch({
+          filterGroups: [
+            {
+              filters: [
+                {
+                  propertyName: 'lastmodifieddate',
```
tomkit, line 469: is the contact property name here diff than others? there's no `hs_` prefix

albertyfwu: correct. i can leave a comment

tomkit (approved):

packages/core/remotes/crm/hubspot/index.ts
albertyfwu: added

Issue #489, opened 2023-03-29
Sync deleted records


albertyfwu:
When records are deleted in hubspot/salesforce, we don’t delete them in our system. Is this important? It looks like it’s possible in Salesforce by including deleted items in query. (and maybe also in hubspot by including archived records)

albertyfwu:
hubspot completed. will do salesforce later.

PR #492, opened 2023-03-29
chore: add new DB columns and types to track deleted_at ts

albertyfwu:

This does not yet actually pull in deleted records from hubspot. It just lays the groundwork to be able to pull it in, by defining the types.

### DIFF
--- a/packages/core/mappers/account.ts
+++ b/packages/core/mappers/account.ts
@@ -52,6 +52,16 @@ export const fromRemoteAccountToDbAccountParams = (
   customerId: string,
   remoteAccount: RemoteAccount
 ) => {
+  const lastModifiedAt =
+    remoteAccount.remoteUpdatedAt || remoteAccount.detectedOrRemoteDeletedAt
+      ? new Date(
+          Math.max(
+            remoteAccount.remoteUpdatedAt?.getTime() || 0,
+            remoteAccount.detectedOrRemoteDeletedAt?.getTime() || 0
+          )
+        )
+      : undefined;
+
   return {
     id: uuidv4(),
     name: remoteAccount.name,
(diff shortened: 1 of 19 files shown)


packages/db/prisma/schema.prisma
```diff
@@ -79,155 +79,165 @@ model Connection {
...
+  customerId                String           @map("customer_id")
+  connectionId              String           @map("connection_id")
+  name                      String?
+  email                     String?
+  isActive                  Boolean?         @map("is_active")
+  remoteCreatedAt           DateTime?        @map("remote_created_at")
+  remoteUpdatedAt           DateTime?        @map("remote_updated_at")
+  remoteWasDeleted          Boolean          @default(false) @map("remote_was_deleted")
+  remoteDeletedAt           DateTime?        @map("remote_deleted_at")
+  detectedOrRemoteDeletedAt DateTime?        @map("detected_or_remote_deleted_at")
```
albertyfwu, line 233: if remote has deleted_at, we use that. if not, we generate a timestamp when we realize the record is deleted

packages/db/prisma/schema.prisma
```diff
@@ -79,155 +79,165 @@ model Connection {
...
+  id                        String           @id @default(uuid())
+  remoteId                  String           @map("remote_id")
+  customerId                String           @map("customer_id")
+  connectionId              String           @map("connection_id")
+  name                      String?
+  email                     String?
+  isActive                  Boolean?         @map("is_active")
+  remoteCreatedAt           DateTime?        @map("remote_created_at")
+  remoteUpdatedAt           DateTime?        @map("remote_updated_at")
+  remoteWasDeleted          Boolean          @default(false) @map("remote_was_deleted")
```
tomkit, line 231: do we use `was` here because we are referring to the remote resource?

packages/db/prisma/schema.prisma
tomkit: i wonder if the precision here (calling out "detected") causes more cognitive overhead in consuming this if developers don't care to distinguish between remote vs our timestamp

tomkit (approved):
left a comment about detectedOrRemoteDeletedAt (we could change down the road), but otherwise lgtm

asdfryan (approved):

albertyfwu:
agreed that the two ts are similar and maybe not necessary. for now, maybe we can just leave it here, since neither column is exposed to the end user. we can remove in a future PR if we want

packages/core/mappers/account.ts
```diff
@@ -52,6 +52,16 @@ export const fromRemoteAccountToDbAccountParams = (
   customerId: string,
   remoteAccount: RemoteAccount
 ) => {
+  const lastModifiedAt =
```
tomkit, line 55: we use `modified` here to differentiate between `update` (non-deleted update only)?

PR #494, opened 2023-03-29
feat: pull in deleted objects from hubspot

albertyfwu:

this actually pulls in deleted objects from hubspot. we cannot use search api, since it doesn't allow filtering on archived. so we have to pull in all the deleted objects all the time (we're hoping that people don't archive that many objects usually). down the road, we might consider using webhook instead

### DIFF
--- a/packages/core/remotes/crm/hubspot/index.ts
+++ b/packages/core/remotes/crm/hubspot/index.ts
@@ -164,6 +164,7 @@ class HubSpotClient extends AbstractCrmRemoteClient {
 
     const passThrough = new PassThrough({ objectMode: true });
 
+    // Accounts
     (async () => {
       let after = undefined;
       do {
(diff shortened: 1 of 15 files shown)


packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -185,17 +186,43 @@ class HubSpotClient extends AbstractCrmRemoteClient {
       passThrough.emit('error', err);
     });
 
+    // Archived Accounts
+    // TODO: this doesn't respect the `updatedAfter` parameter because we can't use the search API for archived accounts
+    (async () => {
+      let after = undefined;
+      do {
+        const currResults: HubspotPaginatedCompanies = await this.#listAccountsFullImpl(after, /* archived */ true);
```
tomkit, line 194: nit: should we just use named parameters if we need to leave a comment?

tomkit (approved):

packages/core/remotes/crm/hubspot/index.ts
albertyfwu: originally i was going to use named parameters, but then it makes it harder to share `listAccountsFullImpl` and `listAccountsIncrementalImpl`.

PR #496, opened 2023-03-29
feat: allow querying for deleted data too

albertyfwu:

allow querying for deleted data by passing in `include_deleted_data: true` in path query params

### DIFF
--- a/openapi/common/components/parameters/query/include_deleted_data.yaml
+++ b/openapi/common/components/parameters/query/include_deleted_data.yaml
@@ -1,4 +1,4 @@
-name: include_deleted_at
+name: include_deleted_data
 in: query
 schema:
   type: boolean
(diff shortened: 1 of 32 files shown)


openapi/common/components/parameters/query/include_deleted_data.yaml
```diff
@@ -0,0 +1,6 @@
+name: include_deleted_at
```
tomkit, line 1: `include_deleted_data` or `include_deleted_at`?

albertyfwu: oops

albertyfwu: fixed

tomkit (approved):

PR #508, opened 2023-03-30
fix: local nextauth

tomkit:
Fixes locally running development environment

## Test Plan

Logged in locally using next-auth

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -17,6 +17,7 @@ cp docker-compose.override.dev.yml docker-compose.override.yml
 
 ```shell
 cp apps/mgmt-ui/.env.sample apps/mgmt-ui/.env
+docker compose up
 
 yarn workspace mgmt-ui dev
 ```
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -16,7 +16,9 @@ cp docker-compose.override.dev.yml docker-compose.override.yml
 ### Running
 
 ```shell
-docker-compose up
+cp apps/mgmt-ui/.env.sample apps/mgmt-ui/.env
+
+yarn workspace mgmt-ui dev
 ```
 
 ### Environment variables
--- a/apps/mgmt-ui/.env.sample
+++ b/apps/mgmt-ui/.env.sample
@@ -1,0 +1,4 @@
+NEXTAUTH_URL=http://localhost:3000
+SUPAGLUE_JWT_SECRET=some-secret
+SUPAGLUE_INTERNAL_TOKEN=some-internal-token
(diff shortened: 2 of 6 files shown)


apps/mgmt-ui/src/pages/index.tsx
```diff
@@ -1,34 +1,22 @@
...
 import { type GetServerSideProps } from 'next';
 import { getServerSession, Session } from 'next-auth';
-import { API_HOST, IS_CLOUD, ORGANIZATION_ID, SG_INTERNAL_TOKEN } from './api';
+import { API_HOST, ORGANIZATION_ID, SG_INTERNAL_TOKEN } from './api';
 
 export const getServerSideProps: GetServerSideProps = async ({ req, res }) => {
   let session: Session | null = null;
 
-  let orgId = ORGANIZATION_ID;
+  const orgId = ORGANIZATION_ID;
```
asdfryan, line 9: I think this will break it for cloud. Did you test with cloud = true?

apps/mgmt-ui/.env.sample
```diff
@@ -0,0 +1,4 @@
+NEXTAUTH_URL=http://localhost:3000
+SUPAGLUE_JWT_SECRET=some-secret
+SUPAGLUE_INTERNAL_TOKEN=some-internal-token
+SUPAGLUE_API_HOST=http://localhost:8080
```
lucasmarshall, line 4: We should add to the development docs that you need to copy this

asdfryan (changes requested):
I think your change will break cloud, let me know if this isn't the case.

apps/mgmt-ui/src/pages/index.tsx
tomkit: vscode did this -- it doesn't look like it's getting assigned anywhere in the code block

tomkit: I removed some code in test -- restored

DEVELOPMENT.md
```diff
@@ -16,7 +16,9 @@ cp docker-compose.override.dev.yml docker-compose.override.yml
 ### Running
 
 ```shell
-docker-compose up
+cp apps/mgmt-ui/.env.sample apps/mgmt-ui/.env
```
lucasmarshall, line 19: Still need to run `docker compose up`

tomkit: This is to run the entire stack or just the FE

lucasmarshall: This is for running the full stack

lucasmarshall (approved):

PR #513, opened 2023-03-30
fix: missing phone number on lead

albertyfwu:
Fixes: #### 

[Describe your change here]

## Test Plan

[Describe test plan here]

## Deployment instructions


packages/core/mappers/lead.ts
```diff
@@ -69,14 +69,15 @@ export const fromRemoteLeadToDbLeadParams = (connectionId: string, customerId: s
...
-    connection_id: connectionId,
     customer_id: customerId,
+    connection_id: connectionId,
     lead_source: remoteLead.leadSource,
     title: remoteLead.title,
     company: remoteLead.company,
     first_name: remoteLead.firstName,
     last_name: remoteLead.lastName,
     addresses: remoteLead.addresses,
+    phone_numbers: remoteLead.phoneNumbers,
```
albertyfwu, line 80: this is the only real change. i reordered the other ones to visually compare against another list in `*_service.ts`. We should really have some type safety here when we have time.

asdfryan (approved):

PR #519, opened 2023-03-31
fix: Dependabot alerts

lucasmarshall:
Corrects all current dependabot alerts in the repo.

## Test Plan

Tested locally. The `got` upgrade was particularly risky since it's a couple of major versions behind. The only think that depended on it was `docusaurus` and docs still work fine

## Deployment instructions

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,5 +56,11 @@
     "posthog-node": "^2.5.4",
     "sync-worker": "workspace:^",
     "workspace": "^0.0.1-preview.1"
+  },
+  "resolutions": {
+    "qs": "6.9.7",
+    "webpack": "5.76.0",
+    "trim": "0.0.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -56,5 +56,11 @@
     "posthog-node": "^2.5.4",
     "sync-worker": "workspace:^",
     "workspace": "^0.0.1-preview.1"
+  },
+  "resolutions": {
```
albertyfwu, line 60: do a lot of libraries depend on these? have we tried bumping those libraries instead of using resolutions?

albertyfwu (approved):

package.json
lucasmarshall: I couldn't find any direct dependencies that we would be able to bump without using resolutions.

Issue #500, opened 2023-03-30
Ability to delete/update applications


george-xing:

albertyfwu:
delete is not trivial because we need to do cleanup of temporal, etc.

PR #521, opened 2023-03-31
feat: allow updating and deleting application

albertyfwu:
Fixes: https://github.com/supaglue-labs/supaglue/issues/500

BE only

## Test Plan

[Describe test plan here]

## Deployment instructions


apps/sync-worker/services/application_service.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Application } from '@supaglue/types';
+
+export class ApplicationService {
+  #prisma: PrismaClient;
+
+  constructor(prisma: PrismaClient) {
+    this.#prisma = prisma;
+  }
+
+  public async getById(id: string): Promise<Application> {
```
asdfryan, line 13: hmm why not use the one in core?

asdfryan (approved):

PR #527, opened 2023-04-01
feat: allow update name and delete for application

albertyfwu:
Fixes: #500 

image

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/mgmt-ui/src/components/ApplicationMenu/index.tsx
+++ b/apps/mgmt-ui/src/components/ApplicationMenu/index.tsx
@@ -22,7 +22,7 @@ import * as React from 'react';
 export default function ApplicationMenu() {
   const router = useRouter();
   const { applications = [], mutate } = useApplications();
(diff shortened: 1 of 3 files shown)


apps/mgmt-ui/src/components/ApplicationMenu/index.tsx
```diff
@@ -38,6 +41,21 @@ export default function ApplicationMenu() {
     await router.push(`/applications/${newApplication.id}`);
   };
 
+  const onUpdateApplicationName = async (id: string, name: string) => {
+    handleClose();
+    const newApplication = await updateApplicationName(id, name);
+    await mutate(applications.map((application) => (application.id === id ? newApplication : application)));
```
asdfryan, line 47: Do you also need to update the active application?

albertyfwu: yes good catch

asdfryan (approved):

PR #544, opened 2023-04-03
fix: upsert application to avoid race

asdfryan:
Fixes: #### 

We should upsert application to avoid race instead of creating since we have a unique constraint on `name, orgId`

## Test Plan

Tested locally

## Deployment instructions

### DIFF
--- a/apps/api/services/application_service.ts
+++ b/apps/api/services/application_service.ts
@@ -2,7 +2,7 @@ import { NotFoundError } from '@supaglue/core/errors';
 import { cryptoHash, generateApiKey } from '@supaglue/core/lib/crypt';
 import { fromApplicationModel } from '@supaglue/core/mappers/application';
 import type { PrismaClient } from '@supaglue/db';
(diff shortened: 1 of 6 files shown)


apps/api/services/application_service.ts
```diff
@@ -63,17 +63,25 @@ export class ApplicationService {
     return applications.map(fromApplicationModel);
   }
 
-  public async create(createParams: ApplicationCreateParams): Promise<Application> {
-    const createdApplication = await this.#prisma.application.create({
-      data: {
+  public async upsert(createParams: ApplicationUpsertParams): Promise<Application> {
```
albertyfwu, line 66: feels like createIfNotExist is better, since we already have an `update` endpoint, but don't feel strongly

albertyfwu (approved):

apps/api/services/application_service.ts
asdfryan: I prefer upsert because we end up using the payload in the FE

PR #556, opened 2023-04-05
fix(helm): allow prisma params to be configured

lucasmarshall:


helm/values.yaml
```diff
@@ -21,6 +21,10 @@ syncWorker:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  db:
+    parameters:
+      connectionLimit: 0 # prisma defaults to num_cpu * 2 + 1
+      poolTimeout: 10 # primsa default is 10 secs
```
tomkit, line 27: did you mean to reverse these two values?

tomkit (approved):

Issue #422, opened 2023-03-23
Index DB columns that are relevant for populating associations, filters, etc.


tomkit:
We don't apply any indexes on DB for filters, populating associations, etc.

We might want to do this for performance reasons (insofar as it prevents toppling over).

albertyfwu:
re-opening to investigate partitions

albertyfwu:
looks fine for now. we can look into partitions in the future for cloud

PR #562, opened 2023-04-05
chore: db indexes for performance

albertyfwu:
Fixes: https://github.com/supaglue-labs/supaglue/issues/422

These are some indexes / query tweaks that in theory may help with:
1. upserting from temp table to main table
1. populating associations
1. list queries against our API sorted by `last_modified_date`

Need to generate migration, test these out, etc., consider other indexes, partitions, etc.

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/services/common_models/event_service.ts
+++ b/packages/core/services/common_models/event_service.ts
@@ -206,12 +206,17 @@ export class EventService extends CommonModelBaseService {
     );
   }
 
-  public async updateDanglingOwners(connectionId: string): Promise<void> {
+  public async updateDanglingOwners(connectionId: string, startingLastModifiedAt: Date): Promise<void> {
     const eventsTable = COMMON_MODEL_DB_TABLES['events'];
     const usersTable = COMMON_MODEL_DB_TABLES['users'];
 
     await this.prisma.crmEvent.updateMany({
       where: {
+        // Only update events for the given connection and that have been updated since the last sync (to be more efficient).
+        connectionId,
+        lastModifiedAt: {
+          gt: startingLastModifiedAt,
+        },
         remoteOwnerId: null,
         ownerId: {
           not: null,
@@ -228,6 +233,7 @@ export class EventService extends CommonModelBaseService {
       FROM ${usersTable} u
       WHERE
         c.connection_id = '${connectionId}'
+        AND c.last_modified_at > '${startingLastModifiedAt.toISOString()}'
         AND c.connection_id = u.connection_id
         AND c.owner_id IS NULL
         AND c._remote_owner_id IS NOT NULL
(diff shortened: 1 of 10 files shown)


tomkit (approved):

packages/sync-workflows/activities/populate_associations.ts
```diff
@@ -12,45 +14,54 @@ export function createPopulateAssociations(
...
   eventService: EventService
 ) {
-  return async function populateAssociations({ connectionId }: PopulateAssociationsArgs) {
+  return async function populateAssociations({
+    connectionId,
+    originalMaxLastModifiedAtMsMap,
+  }: PopulateAssociationsArgs) {
     // TODO: Parallelize / optimize? Keeping serial for now to be safe. Each command should run pretty quickly anyway.
-    await contactService.updateDanglingAccounts(connectionId);
+    await contactService.updateDanglingAccounts(connectionId, new Date(originalMaxLastModifiedAtMsMap['contact']));
```
asdfryan, line 22: Shouldn't this be `originalMaxLastModifiedAtMsMap['account']` instead?

packages/db/prisma/schema.prisma
```diff
@@ -144,6 +145,7 @@ model CrmContact {
   events                    CrmEvent[]
 
   @@unique([connectionId, remoteId])
+  @@index([connectionId, lastModifiedAt(sort: Asc)]) // for incremental syncs hitting our API // TODO: consider adding index on `remoteWasDeleted`
```
asdfryan, line 148: We may want to add indices on the association fields

packages/core/services/common_models/event_service.ts
```diff
@@ -286,19 +303,25 @@ export class EventService extends CommonModelBaseService {
       FROM ${contactsTable} u
       WHERE
         c.connection_id = '${connectionId}'
+        AND c.last_modified_at > '${startingLastModifiedAt.toISOString()}'
```
asdfryan, line 306: I haven't thought about this, but `startingLastModifiedAt` is for events right? Should this instead be for contacts instead?

albertyfwu: it should be for `events`. We only need to populate associations for events that we know have been updated in this sync cycle

packages/db/prisma/schema.prisma
albertyfwu: let's talk offline

packages/sync-workflows/activities/populate_associations.ts
albertyfwu: no it should be for the table on which we're doing the updates

PR #569, opened 2023-04-05
fix: paginate upsert from temp table

albertyfwu:
Fixes: #### 

[Describe your change here]

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/services/common_models/base_service.ts
+++ b/packages/core/services/common_models/base_service.ts
@@ -30,7 +30,8 @@ export abstract class CommonModelBaseService {
     tempTable: string,
     columnsWithoutId: string[],
     mapper: (connectionId: string, customerId: string, remoteCommonModel: T) => Record<string, any>,
-    lastModifiedAtGetter: (remoteCommonModel: T) => Date | null
+    lastModifiedAtGetter: (remoteCommonModel: T) => Date | null,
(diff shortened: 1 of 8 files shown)


lucasmarshall (approved):

packages/core/services/common_models/base_service.ts
```diff
@@ -96,24 +97,37 @@ export abstract class CommonModelBaseService {
...
 
-      // IMPORTANT: we need to use DISTINCT ON because we may have multiple records with the same remote_id
-      // For example, hubspot will return the same record twice when querying for `archived: true` if
-      // the record was archived, restored, and archived again.
-      // TODO: This may have performance implications. We should look into this later.
-      // https://github.com/supaglue-labs/supaglue/issues/497
-      const result = await client.query(`INSERT INTO ${table}
-SELECT DISTINCT ON (remote_id) * FROM ${tempTable}
+      // Paginate
+      const tempTableRowCountResult = await client.query(`SELECT COUNT(*) AS C FROM ${tempTable}`);
```
tomkit: Can we accumulate this from the stream above?

albertyfwu: yea true. let me do this in a follow-up PR so i can at least get this in first

PR #588, opened 2023-04-07
chore: endpoint to remove orphaned WFs

albertyfwu:
[Describe your change here]

## Test Plan

[Describe test plan here]

## Deployment instructions


lucasmarshall (approved):

openapi/mgmt/openapi.yaml
```diff
@@ -43,6 +43,8 @@ paths:
     $ref: paths/sync_history.yaml
   '/sync-info':
     $ref: paths/sync_info.yaml
+  '/sync/_manually_clean_up_orphaned_temporal_syncs':
```
asdfryan, line 46: This will generate openapi docs for this. Do we really want to expose this in our public API documentation?

asdfryan: IMO we shouldn't, and we should just disable the openapi middleware for this endpoint.

albertyfwu: done

PR #589, opened 2023-04-07
chore: manual endpoint for fixing temporal syncs

albertyfwu:
The previous PR would delete orphaned WFs/schedules in Temporal.
This PR improves the endpoint to also spin up new Temporal schedules if somehow it was manually deleted by someone (even though the sync exists in Postgres).

Also moved endpoint to /internal

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/api/routes/mgmt/v1/index.ts
+++ b/apps/api/routes/mgmt/v1/index.ts
@@ -15,8 +15,8 @@ export default function init(app: Router): void {
   const v1Router = Router();
 
   v1Router.use(apiKeyHeaderMiddleware);
-  v1Router.post('/_manually_clean_up_orphaned_temporal_syncs', async (req, res) => {
-    const result = await connectionAndSyncService.manuallyCleanUpOrphanedTemporalSyncs();
+  v1Router.post('/_manually_fix_temporal_syncs', async (req, res) => {
+    const result = await connectionAndSyncService.manuallyFixTemporalSyncs();
     return res.status(200).send(snakecaseKeys(result));
   });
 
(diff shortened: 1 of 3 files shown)


apps/api/routes/mgmt/v1/index.ts
```diff
@@ -15,8 +15,8 @@ export default function init(app: Router): void {
   const v1Router = Router();
 
   v1Router.use(apiKeyHeaderMiddleware);
-  v1Router.post('/_manually_clean_up_orphaned_temporal_syncs', async (req, res) => {
-    const result = await connectionAndSyncService.manuallyCleanUpOrphanedTemporalSyncs();
+  v1Router.post('/_manually_fix_temporal_syncs', async (req, res) => {
```
lucasmarshall, line 18: Can we rename to `_manually_fix_syncs`? temporal is an implementation detail

albertyfwu: fixed

apps/api/services/connection_and_sync_service.ts
```diff
@@ -249,12 +260,97 @@ export class ConnectionAndSyncService {
...
+        throw new Error('Unexpected error: connection not found');
+      }
+      const integration = integrations.find((integration) => integration.id === connection.integrationId);
+      if (!integration) {
+        throw new Error('Unexpected error: integration not found');
+      }
+      const justCreated = await this.#createTemporalSyncIfNotExist(
+        syncId,
+        connection,
+        integration.config.sync.periodMs ?? FIFTEEN_MINUTES_MS
```
tomkit, line 289: nit: I think the default we use in the mgmt-ui is 1 hour

tomkit (approved):

PR #607, opened 2023-04-11
fix: quickstart version

asdfryan:


albertyfwu (approved):
we should fix for 0.7.3 as well

PR #622, opened 2023-04-12
fix: salesforce refresh

asdfryan:
This fixes a few issues:
- the main issue: persist connection loginUrl so we can use the correct url when refreshing access tokens
- don't infinitely retry when refreshing a token fails
- introduce a timeout when fetching (this is unrelated, but good to do anyway)

### DIFF
--- a/packages/core/remotes/crm/salesforce/index.ts
+++ b/packages/core/remotes/crm/salesforce/index.ts
@@ -298,7 +298,9 @@ class SalesforceClient extends AbstractCrmRemoteClient {
     const interval = 1000; // TODO: make configurable
 
     while (startTime + timeout > Date.now()) {
-      const { state } = await poll();
+      const pollResponse = await poll();
+      logger.info(`poll response: `, pollResponse);
+      const { state } = pollResponse;
       switch (state) {
         case 'Open':
           throw new Error('job has not been started');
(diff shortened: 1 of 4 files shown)


packages/core/remotes/crm/salesforce/index.ts
```diff
@@ -253,7 +263,16 @@ class SalesforceClient extends AbstractCrmRemoteClient {
       // CSV download for bulk 2.0 with large loads and also more future control of rate limits,
       // etc.
       if (response.status === 401) {
-        await this.#refreshAccessToken();
+        try {
+          await this.#refreshAccessToken();
+        } catch (e: any) {
+          logger.error(e);
+          if (e.name === 'ERROR_HTTP_429') {
```
albertyfwu, line 270: is the error name exactly this, or should this be `startsWith`? Can you give an example and leave a note to point to where this error gets generated in the jsforce library? Also, can we leave a comment/GH issue to possibly move off of jsforce here?

packages/types/connection.ts
```diff
@@ -2,12 +2,15 @@ import type { CRMProviderName } from './crm';
 
 export type ConnectionStatus = 'available' | 'added' | 'authorized' | 'callable';
 
+// TODO: Bifurcate salesforce vs hubspot
 export type ConnectionCredentialsDecrypted = {
   type: string;
   accessToken: string;
   refreshToken: string;
   expiresAt: string | null; // null means unknown expiry time
+  // Needed for salesforce only
```
albertyfwu, line 11: create GH issue to bifurcate types

albertyfwu (approved):

lucasmarshall (approved):

packages/core/remotes/crm/salesforce/index.ts
asdfryan: https://github.com/jsforce/jsforce/blob/master/lib/oauth2.js#L197

PR #633, opened 2023-04-14
docs: update salesforce reference docs

george-xing:
[Describe your change here]

## Test Plan

[Describe test plan here]

## Deployment instructions


docs/docs/references/salesforce.md
```diff
@@ -1,3 +1,63 @@
...
-*Docs to be added...*
+import BrowserWindow from '@site/src/components/BrowserWindow';
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+To connect to your customers' Salesforce instances, you'll need to update the redirect URL to point to Supaglue and fetch the API access credentials in your [Salesforce developer account](https://developer.salesforce.com).
+
+:::caution
+
+1. Salesforce API access is limited to Enterprise, Unlimited, and Performance Editions of Salesforce, as well as Professional Editions with the paid API add-on. See Salesforce documentation for more information.
```
lucasmarshall, line 11: Should clarify that this is the customer's customer's Salesforce account we are talking about here. Given that, it may not belong in this section, since this is about our customer configuring their app in their developer account.

lucasmarshall (changes requested):

docs/docs/references/salesforce.md
george-xing: good point, I'll remove for now

lucasmarshall (approved):

PR #653, opened 2023-04-18
fix: mgmt-ui deployment

lucasmarshall:


asdfryan (approved):

.github/workflows/deploy_management_ui.yml
```diff
@@ -15,8 +15,8 @@ jobs:
       - name: Install Dependencies
         run: yarn install
       - name: Pull Vercel Environment Information
-        run: yarn workspace mgmt-ui vercel pull --yes --environment=production --token=${{ secrets.VERCEL_TOKEN }}
```
tomkit, line 18: What was the issue here, were there some artifacts from another github action or something?

lucasmarshall: Vercel expects to be run from the root of the repo

PR #654, opened 2023-04-18
fix: sentry reporting for API and sync workers

asdfryan:
Fixes sentry reporting:
- Our logger stream to sentry was only working the very first time -- this PR fixes this
- Temporal activity/workflow errors are reporting at WARNING level. To get around this, I'm logging error every time a sync fails.

## Test Plan

Tested locally

[Describe test plan here]

## Deployment instructions


packages/core/lib/logger.ts
```diff
@@ -15,26 +14,12 @@ if (process.env.SUPAGLUE_PRETTY_LOGS) {
...
-
-      const { err, msg, level } = obj;
-
-      // warning is 40, error is 50, fatal is 60
-      if (level >= 40) {
-        if (err) {
-          const mappedErr = new Error(err.message);
-          mappedErr.stack = err.stack;
-          Sentry.captureException(mappedErr);
-          return;
```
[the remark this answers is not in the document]
asdfryan: possibly, but let's just use the standard thing for now

packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -836,7 +836,7 @@ const retryWhenRateLimited = async <Args extends any[], Return>(
     try {
       return await operation(...parameters);
     } catch (e: any) {
-      logger.error(e, 'Error encountered');
+      logger.error(e, `Error encountered: ${e.message}`);
```
albertyfwu, line 839: won't the error message be included by pino? why do we need to interpolate it here?

albertyfwu (approved):

packages/sync-workflows/activities/log_sync_finish.ts
```diff
@@ -1,16 +1,47 @@
...
   }: {
+    syncId: string;
+    connectionId: string;
     historyId: string;
     status: SyncHistoryStatus;
     errorMessage?: string;
   }) {
     await syncHistoryService.logFinish({ historyId, status, errorMessage });
+    if (status === 'FAILURE') {
+      logger.error(
```
tomkit, line 23: nit: can we log the message in a fixed way and log the dynamic/contextual part as part of the metadata argument to make it easier to search?

packages/core/remotes/crm/hubspot/index.ts
tomkit: +1

PR #666, opened 2023-04-19
chore: fix up types to remove external/internal association ids

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/types/crm/account.ts
+++ b/packages/types/crm/account.ts
@@ -1,16 +1,7 @@
-import type {
-  Address,
-  BaseCrmModel,
-  BaseCrmModelNonRemoteParams,
-  BaseCrmModelRemoteOnlyParams,
-  CustomFields,
-  LifecycleStage,
-  PhoneNumber,
-  User,
-} from '..';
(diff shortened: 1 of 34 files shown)


asdfryan (approved):

asdfryan (changes requested):
looks like build is failing

packages/types/crm/account.ts
```diff
@@ -1,43 +1,22 @@
...
   lifecycleStage: LifecycleStage | null;
-};
 
-export type Account = BaseAccount &
-  BaseCrmModelNonRemoteParams & {
-    ownerId: string | null;
-    owner?: User;
-    // TODO: Support remote data
-  };
+  ownerId: string | null;
```
asdfryan, line 14: Are we getting rid of the associations?

albertyfwu: i'd prefer not to address `expand` for now. we can implement that in another PR.

asdfryan (approved):

PR #671, opened 2023-04-19
chore: postgres config

asdfryan:
## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/types/destination.ts
+++ b/packages/types/destination.ts
@@ -8,46 +8,47 @@ type BaseDestinationUpdateParams = BaseDestination;
 
 export type S3Destination = BaseDestination & {
   type: 's3';
-  // TODO: encryption
-  config: {
-    region: string; // us-west-2
-    bucket: string;
-    accessKeyId: string;
-    secretAccessKey: string;
-  };
+  // TODO(670): encryption
+  config: S3Config;
 };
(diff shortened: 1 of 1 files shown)


packages/types/destination.ts
```diff
@@ -8,46 +8,47 @@ type BaseDestinationUpdateParams = BaseDestination;
...
-  config: object; // TODO: define this
+  config: PostgresConfig;
 };
 export type PostgresDestinationUpdateParams = BaseDestinationUpdateParams & {
   type: 'postgres';
-  config: object; // TODO: define this
+  config: PostgresConfig;
+};
+
+export type PostgresConfig = {
```
albertyfwu, line 45: where does the customer specify the schema we should write to? or should we use schema to encode `crm`, `hris`, etc.?

asdfryan: We hadn't decided on the shape yet -- as far as I know it's going to be not configurable anyway (either common model, or json) so I'm not sure we'd expose it anyway

albertyfwu: by schema i mean postgres `schema`, like `public`

asdfryan: I see. I guess I can add this (future PR)

albertyfwu (approved):

PR #673, opened 2023-04-20
chore: add original stack trace for failed syncs

tomkit:
- Convert logger obj to include context
- Make the error message fixed
- Add stack trace to logger line

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/sync-workflows/workflows/run_sync.ts
+++ b/packages/sync-workflows/workflows/run_sync.ts
@@ -74,7 +74,7 @@ export async function runSync({ syncId, connectionId }: RunSyncArgs): Promise<vo
         break;
     }
   } catch (err: any) {
-    const errorMessage = getErrorMessage(err);
+    const { message: errorMessage, stack: errorStack } = getErrorMessageStack(err);
     await Promise.all(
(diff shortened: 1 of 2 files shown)


packages/sync-workflows/workflows/run_sync.ts
```diff
@@ -74,7 +74,7 @@ export async function runSync({ syncId, connectionId }: RunSyncArgs): Promise<vo
         break;
     }
   } catch (err: any) {
-    const errorMessage = getErrorMessage(err);
+    const { message: errorMessage, stack: errorStack } = getErrorMessageStack(err);
     await Promise.all(
       CRM_COMMON_MODELS.map(async (commonModel) => {
         await logSyncFinish({
```
albertyfwu, line 80: why aren't we just literally passing the error in? i think we are doing too much conversion/mapping here

asdfryan: We can't pass the error in because of how temporal wraps errors. We won't see the error message.

asdfryan:
I pushed a fix

asdfryan (approved):

PR #676, opened 2023-04-20
fix: sentry stack trace

asdfryan:
This PR makes it so that we intercept Warnings with error payloads and send them as errors to Sentry.

This works as a workaround because:
- Temporal logs uncaught errors as warnings (with stack trace etc)
- By the time the sync has failed, the activity failure object does not have the stack trace in its original form

## Test Plan

[Describe test plan here]

## Deployment instructions


albertyfwu (approved):
can you remove `pino-sentry` as a dependency too?

albertyfwu (approved):

PR #677, opened 2023-04-20
fix: return an 401 instead of a 404 on unknown api key

lucasmarshall:
Also don't log api key


albertyfwu (approved):

albertyfwu (approved):
is there some way to log the applicationId when we DO get a valid api key?

asdfryan (approved):

lucasmarshall:
> is there some way to log the applicationId when we DO get a valid api key?

Yes. Will do in another PR

PR #681, opened 2023-04-20
chore: add postgres schema

asdfryan:
ability to specify postgres schema

## Test Plan

[Describe test plan here]

## Deployment instructions

[Add any special deployment instructions here]

PR #682, opened 2023-04-20
feat: write to postgres destination

albertyfwu:
Write records to postgres destination. Currently, it uses a hard-coded postgres schema. When https://github.com/supaglue-labs/supaglue/pull/681 lands, we can pull schema from the config instead.

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/openapi/crm/components/schemas/objects/addresses.yaml
+++ b/openapi/crm/components/schemas/objects/addresses.yaml
@@ -26,11 +26,11 @@ items:
       type: string
       nullable: true
       example: CA
-    street1:
+    street_1:
       type: string
       nullable: true
       example: 525 Brannan
-    street2:
+    street_2:
       type: string
       nullable: true
       example: ~
@@ -40,12 +40,13 @@ items:
     - country
     - postal_code
     - state
-    - street1
+    - street_1
+    - street_2
 example:
   - address_type: shipping
     city: San Francisco
     country: US
     postal_code: '94107'
     state: CA
-    street1: 525 Brannan
-    street2: ~
+    street_1: 525 Brannan
+    street_2: ~
--- a/packages/core/destination_writers/postgres.ts
+++ b/packages/core/destination_writers/postgres.ts
@@ -1,0 +1,389 @@
+import { ConnectionSafeAny, CRMCommonModel, PostgresDestination } from '@supaglue/types';
+import { stringify } from 'csv-stringify';
+import { Pool, PoolClient } from 'pg';
+import { from as copyFrom } from 'pg-copy-streams';
+import { Readable, Transform } from 'stream';
+import { pipeline } from 'stream/promises';
+import { keysOfSnakecasedAccountWithTenant } from '../keys/account';
+import { keysOfSnakecasedContactWithTenant } from '../keys/contact';
+import { keysOfSnakecasedEventWithTenant } from '../keys/event';
+import { keysOfSnakecasedLeadWithTenant } from '../keys/lead';
+import { keysOfSnakecasedOpportunityWithTenant } from '../keys/opportunity';
+import { keysOfSnakecasedUserWithTenant } from '../keys/user';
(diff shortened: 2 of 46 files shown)


openapi/crm/components/schemas/objects/addresses.yaml
```diff
@@ -26,11 +26,11 @@ items:
       type: string
       nullable: true
       example: CA
-    street1:
```
asdfryan, line 29: A couple weeks ago I changed it to this because if you convert this to camelcase and then back to snakecase, it remains street1 (doesn't work for street_1).

packages/core/destination_writers/postgres.ts
```diff
@@ -0,0 +1,333 @@
...
+
+  public override async writeObjects(
+    { id: connectionId, providerName, customerId }: ConnectionSafeAny,
+    commonModelType: CRMCommonModelType,
+    inputStream: Readable,
+    onUpsertBatchCompletion: (offset: number, numRecords: number) => void
+  ): Promise<WriteCommonModelsResult> {
+    const childLogger = logger.child({ connectionId, providerName, customerId, commonModelType });
+
+    const schema = 'synced'; // TODO: replace with setting from PostgresDestination when it's available
```
asdfryan, line 53: it's available now

packages/core/destination_writers/postgres.ts
```diff
@@ -0,0 +1,333 @@
...
+    const table = tableNamesByCommonModelType[commonModelType];
+    const qualifiedTable = `${schema}.${table}`;
+    const tempTable = `temp_${table}`;
```
asdfryan, line 56: Is it ever possible for 2 syncs to be running simultaneously for the same common model?

openapi/crm/components/schemas/objects/addresses.yaml
albertyfwu: we have explicit mapper now

packages/core/destination_writers/postgres.ts
```diff
@@ -0,0 +1,333 @@
...
+CREATE TABLE IF NOT EXISTS "${schema}"."crm_events" (
+  "provider_name" TEXT NOT NULL,
+  "customer_id" TEXT NOT NULL,
+  "remote_id" TEXT NOT NULL,
+  "type" TEXT,
+  "subject" TEXT,
+  "content" TEXT,
+  "start_time" TIMESTAMP(3),
+  "end_time" TIMESTAMP(3),
+  "remote_created_at" TIMESTAMP(3),
```
asdfryan, line 319: I wonder if we can just get rid of all the `remote_` prefixes

packages/core/destination_writers/postgres.ts
albertyfwu: no, because uniquenes of wokrflow ids in temporal,  and also temp tables are unique to a connection, so it doesn't matter

packages/sync-workflows/activities/import_records.ts
```diff
@@ -39,6 +30,7 @@ export function createImportRecords(
   }: ImportRecordsArgs): Promise<ImportRecordsResult> {
     const connection = await connectionService.getSafeById(connectionId);
     const client = await remoteService.getCrmRemoteClient(connectionId);
+    const writer = await destinationService.getWriterByApplicationId(connection.applicationId);
```
asdfryan, line 33: What happens if you create a connection without any destinations?

packages/core/destination_writers/postgres.ts
albertyfwu: yes, next PR. don't wnat to do additional type chnages here

packages/sync-workflows/activities/import_records.ts
albertyfwu: the sync will fail

asdfryan (approved):

PR #684, opened 2023-04-21
fix: validate sync period ms. fix swr refetch

tomkit:
- Don't allow sync period ms below 1 minute

Separately, when adding validation, I noticed sometimes the sync period shown changed non-deterministically

- Update react hook to always set values when integration is fetched
- Fix `mutate` to not have potentially contain duplicated entry for the case of updates
- Use `optimisticData` for revalidation so value is the latest one saved

Fixes: #### 

Screenshot 2023-04-21 at 6 07 36 PM
Screenshot 2023-04-21 at 6 07 43 PM
Screenshot 2023-04-21 at 6 07 49 PM

[Describe your change here]

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/mgmt-ui/src/components/configuration/IntegrationDetailsPanel.tsx
+++ b/apps/mgmt-ui/src/components/configuration/IntegrationDetailsPanel.tsx
@@ -11,10 +11,16 @@ import { useEffect, useState } from 'react';
 import Spinner from '../Spinner';
 import { integrationCardsInfo } from './IntegrationTabPanelContainer';
 
+const ONE_HOUR_SECONDS = 60 * 60;
+
 export type IntegrationDetailsPanelProps = {
   category: IntegrationCategory;
   providerName: CRMProviderName;
   isLoading: boolean;
+};
+
+const isSyncPeriodSecsValid = (syncPeriodSecs: number) => {
+  return syncPeriodSecs < 60 ? false : true;
 };
 
 export default function IntegrationDetailsPanel({ providerName, category, isLoading }: IntegrationDetailsPanelProps) {
@@ -23,6 +29,7 @@ export default function IntegrationDetailsPanel({ providerName, category, isLoad
   const [clientSecret, setClientSecret] = useState<string>('');
   const [oauthScopes, setOauthScopes] = useState<string>('');
   const [syncPeriodSecs, setSyncPeriodSecs] = useState<number | undefined>();
+  const [isFormValid, setIsFormValid] = useState<boolean>(true);
   const router = useRouter();
 
   const { integrations: existingIntegrations = [], mutate } = useIntegrations();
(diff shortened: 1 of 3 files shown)


apps/mgmt-ui/src/components/configuration/IntegrationDetailsPanel.tsx
```diff
@@ -177,9 +188,18 @@ export default function IntegrationDetailsPanel({ providerName, category, isLoad
             </Button>
             <Button
               variant="contained"
+              disabled={!isFormValid}
               onClick={async () => {
                 const newIntegration = await createOrUpdateIntegration();
-                mutate([...existingIntegrations, newIntegration], false);
```
tomkit, line 182: this had duplicated entries with the same integration.id if updating an existing integration

apps/mgmt-ui/src/components/configuration/IntegrationDetailsPanel.tsx
```diff
@@ -36,18 +43,15 @@ export default function IntegrationDetailsPanel({ providerName, category, isLoad
...
-    if (!clientId) {
-      setClientId(integration?.config?.oauth?.credentials?.oauthClientId ?? '');
-    }
-    if (!clientSecret) {
-      setClientSecret(integration?.config?.oauth?.credentials?.oauthClientSecret ?? '');
-    }
-    if (!oauthScopes) {
-      setOauthScopes(integration?.config?.oauth?.oauthScopes?.join(',') ?? '');
-    }
-    if (!syncPeriodSecs) {
```
tomkit, line 48: if `[integration?.id]` is set we should just set what was fetched. if you navigate away and refocus the tab, `[integration?.id]` will still be the same value so it doesn't cause an issue

apps/mgmt-ui/src/components/configuration/IntegrationDetailsPanel.tsx
```diff
@@ -159,8 +163,15 @@ export default function IntegrationDetailsPanel({ providerName, category, isLoad
             label="Sync every (in seconds)"
             variant="outlined"
             type="number"
+            helperText="Value needs to be 60 seconds or greater."
+            error={syncPeriodSecs === undefined || !isSyncPeriodSecsValid(syncPeriodSecs)}
             onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
-              setSyncPeriodSecs(parseInt(event.target.value, 10));
+              let value: number | undefined = parseInt(event.target.value, 10);
+              if (Number.isNaN(value)) {
```
tomkit, line 170: When this input field is emptied out, it's `undefined`. Setting it to something like `0` caused some issue with the component being controlled/uncontrolled so the placeholder text was overlapping the value

apps/mgmt-ui/src/components/configuration/IntegrationDetailsPanel.tsx
```diff
@@ -11,18 +11,25 @@ import { useEffect, useState } from 'react';
...
+const ONE_HOUR_SECONDS = 60 * 60;
+
 export type IntegrationDetailsPanelProps = {
   category: IntegrationCategory;
   providerName: CRMProviderName;
   isLoading: boolean;
 };
 
+const isSyncPeriodSecsValid = (syncPeriodSecs: number) => {
+  return syncPeriodSecs < 60 ? false : true;
```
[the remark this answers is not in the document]
tomkit: Ryan uses 1 min today for testing

The Temporal Schedule will just delay any to-be-scheduled workflows rather than cause backpressure

lucasmarshall (approved):

PR #690, opened 2023-04-21
chore: remove remote_* references, detected or deleted at

asdfryan:


packages/core/remotes/crm/hubspot/mappers.ts
```diff
@@ -141,7 +140,7 @@ export const fromHubSpotContactToContact = ({
       : [];
 
   return {
-    remoteId: id,
+    id: id,
```
albertyfwu, line 143: nit: replace all `id: id` across all files with just `id`

albertyfwu (approved):

Issue #683, opened 2023-04-21
Lead search


da-wallace:
Similar to the `contact` api. The ability to search for `leads` by email address. This would just be duplicating the contact functions over to leads.

PR #694, opened 2023-04-21
feat: lead search by email

asdfryan:
Fixes: #683

[Describe your change here]

## Test Plan

Tested locally

image

## Deployment instructions

### DIFF
--- a/packages/core/services/common_models/lead_service.ts
+++ b/packages/core/services/common_models/lead_service.ts
@@ -3,9 +3,11 @@ import type {
   GetParams,
   Lead,
   LeadCreateParams,
+  LeadFilters,
   LeadUpdateParams,
   ListInternalParams,
   PaginatedResult,
+  PaginationInternalParams,
 } from '@supaglue/types';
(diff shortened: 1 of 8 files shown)


packages/core/services/common_models/lead_service.ts
```diff
@@ -42,6 +44,34 @@ export class LeadService extends CommonModelBaseService {
...
+            ? {
+                array_contains: [{ emailAddress: filters.emailAddress.value }],
+              }
+            : undefined,
+      },
+      orderBy: {
+        id: 'asc',
+      },
+    });
+    const results = models.map((model) => fromLeadModel(model));
```
albertyfwu, line 68: nit: `models.map(fromLeadModel)`

albertyfwu (approved):

packages/core/services/common_models/lead_service.ts
asdfryan: Won't work because fromLeadModel as an optional 2nd param

PR #700, opened 2023-04-22
fix: turborepo cache for mgmt-ui builds

lucasmarshall:


turbo.json
```diff
@@ -24,6 +24,9 @@
       ],
       "outputs": ["build/**"]
     },
+    "mgmt-ui#build": {
```
tomkit, line 27: what does `#build` denote?

tomkit (approved):

turbo.json
lucasmarshall: The build target

PR #723, opened 2023-04-25
feat(mgmt-ui): Add notification when saving

asdfryan:
Fixes: #722

## Test Plan

Tested locally

https://www.loom.com/share/c184fb000f03484ea3eeb050d8667530

## Deployment instructions

### DIFF
--- a/apps/mgmt-ui/src/client.ts
+++ b/apps/mgmt-ui/src/client.ts
@@ -54,7 +54,7 @@ export async function createOrUpdateWebhook(applicationId: string, data: Webhook
 
 export async function deleteWebhook(applicationId: string) {
   await fetch(`/api/internal/webhook/delete`, {
-    method: 'DELETE',
+    method: 'POST',
     headers: {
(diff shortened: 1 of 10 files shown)


apps/mgmt-ui/src/client.ts
```diff
@@ -52,9 +52,9 @@ export async function createOrUpdateWebhook(applicationId: string, data: Webhook
   return r;
 }
 
-export async function deleteWebhook(applicationId: string): Promise<void> {
+export async function deleteWebhook(applicationId: string) {
   await fetch(`/api/internal/webhook/delete`, {
-    method: 'POST',
```
tomkit, line 57: I originally considered 'DELETE', but since we're not deleting the actual resource, but operating on the "application" resource, I made it a POST (it currently behaves more like RPC)

asdfryan: sure

tomkit (approved):

PR #725, opened 2023-04-25
chore: increase heartbeat timeout for populate associations

asdfryan:
This is because `populateAssociations` happens in 1 step per model, so if there are many records, it may time out.


tomkit:
If there are failures in the activity with heartbeats, will it fail as soon as that happens or does it need to wait until the heartbeat expires?

asdfryan:
> If there are failures in the activity with heartbeats, will it fail as soon as that happens or does it need to wait until the heartbeat expires?

Yes it will fail as soon as that happens

lucasmarshall (approved):

PR #727, opened 2023-04-25
chore: add context to logs

lucasmarshall:


apps/api/routes/internal/v1/index.ts
```diff
@@ -15,18 +14,13 @@ import syncHistory from './sync_history';
...
 
-const { connectionAndSyncService, prisma } = getDependencyContainer();
+const { prisma } = getDependencyContainer();
 
 export default function init(app: Router): void {
   // application routes should not require application header
   const v1ApplicationRouter = Router();
   v1ApplicationRouter.use(internalMiddleware);
 
-  v1ApplicationRouter.post('/_manually_fix_syncs', async (req, res) => {
```
lucasmarshall, line 25: Moved this to a script

packages/core/lib/logger.ts
```diff
@@ -1,45 +1,41 @@
...
-          const mappedErr = new Error(err.message);
-          mappedErr.stack = err.stack;
-          Sentry.captureException(mappedErr);
-          continue;
-        }
-
-        Sentry.captureMessage(msg);
-      }
-    }
+  const sentryStream = createWriteStream({
```
asdfryan, line 23: discussed offline -- I'm not sure this will do what we want, but testing sgtm

apps/api/routes/internal/v1/index.ts
tomkit: How do you execute it now, ssh onto the machine and run the script?

lucasmarshall: kubectl exec into a pod and run it

asdfryan (approved):

PR #733, opened 2023-04-26
feat: ssl for postgres destinations

lucasmarshall:
This needs some testing


apps/mgmt-ui/src/components/configuration/destination/PostgresDestinationDetailsPane.tsx
```diff
@@ -198,6 +205,18 @@ export default function PostgresDestinationDetailsPanel({ isLoading }: PostgresD
           />
         </Stack>
 
+        <Stack className="gap-2">
+          <Typography variant="subtitle1">SSL</Typography>
+          <Select
```
asdfryan, line 210: Should this be a checkbox instead?

lucasmarshall: There probably should be other options down the line.

asdfryan (commented):
LGTM modulo build failures

apps/mgmt-ui/src/components/configuration/destination/PostgresDestinationDetailsPane.tsx
```diff
@@ -58,6 +65,23 @@ export default function PostgresDestinationDetailsPanel({ isLoading }: PostgresD
...
+    if (!ca) {
+      setCa(destination.config.ca);
+    }
+    if (!cert) {
+      setCert(destination.config.cert);
+    }
+    if (!key) {
+      setKey(destination.config.key);
+    }
+    if (!passphrase) {
```
tomkit, line 80: In another PR I removed these `if()` guards now that the 2nd argument (`dependencies`) for it is working. Having the `if()` guards would mean some values are set, but not others based on the value being truthy/falsy

openapi/mgmt/components/schemas/objects/postgres_config.yaml
```diff
@@ -18,10 +18,41 @@ properties:
...
+    example: password
+
 required:
   - host
   - port
   - database
   - schema
   - user
   - password
+  - sslmode
```
tomkit, line 53: We might want to add a lightweight of differentiating between different types of connection configurations, e.g. not all will require certificates

asdfryan (approved):

PR #747, opened 2023-04-27
feat: hubspot stage and pipeline names

asdfryan:
Fixes: #743

Changes:
- We will return the name of the pipeline and stage in the `pipeline` and `stage` fields for Hubspot instead of their ids (i.e. `Sales Pipeline` instead of `default`)
- For Create/Update, we accept either the ID or the name (we will do the lookup)

## Test Plan

Tested locally -- see screenshots below:

Screenshot 2023-04-27 at 2 33 01 PM
Screenshot 2023-04-27 at 2 33 37 PM
Screenshot 2023-04-27 at 2 33 55 PM
image

## Deployment instructions

### DIFF
--- a/packages/core/remotes/crm/hubspot/index.ts
+++ b/packages/core/remotes/crm/hubspot/index.ts
@@ -113,6 +113,8 @@ const propertiesToFetch = {
 const CONTACT_TO_PRIMARY_COMPANY_ASSOCIATION_ID = 1;
 const OPPORTUNITY_TO_PRIMARY_COMPANY_ASSOCIATION_ID = 5;
 
+export type PipelineStageMapping = Record<string, { label: string; stageIdsToLabels: Record<string, string> }>;
+
 type Credentials = {
   accessToken: string;
   refreshToken: string;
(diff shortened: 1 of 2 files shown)


tomkit (approved):

packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -304,7 +306,28 @@ class HubSpotClient extends AbstractCrmRemoteClient {
     return await this.getAccount(company.id);
   }
 
+  async #getPipelineStageMapping(): Promise<
+    Record<string, { label: string; stageIdsToLabels: Record<string, string> }>
+  > {
+    return await retryWhenRateLimited(async () => {
+      await this.maybeRefreshAccessToken();
+      const response = await this.#client.crm.pipelines.pipelinesApi.getAll('deal');
```
lucasmarshall, line 314: Maybe we should cache this for a while? It won't change often, right?

asdfryan: We do this once a sync -- I think the extra cost incurred is very small IMO.

PR #758, opened 2023-04-28
chore: add instance_url migration

asdfryan:


packages/db/prisma/schema.prisma
```diff
@@ -75,7 +75,9 @@ model Connection {
   credentials   Bytes // encrypted, {type, access_token, refresh_token, expires_at, raw}
   customer      Customer    @relation(fields: [customerId], references: [id], onDelete: Cascade)
   customerId    String      @map("customer_id")
+  // Deprecated
```
lucasmarshall, line 78:
```suggestion
  /// Deprecated
```

https://www.prisma.io/docs/concepts/components/prisma-schema

PR #770, opened 2023-05-01
feat: write raw_data into jsonb column

asdfryan:
This is step 2 of supporting raw data. Remaining tasks:

- [ ] Fetch available properties in hubspot before syncing
- [ ] Support fetching raw data in our API endpoints

## Test Plan

Testing locally
## Deployment instructions


packages/core/remotes/crm/hubspot/mappers.ts
```diff
@@ -225,6 +230,17 @@ export const fromHubspotOwnerToRemoteUser = ({
     remoteWasDeleted: !!archived,
     remoteDeletedAt: null,
     detectedOrRemoteDeletedAt: archived ? new Date() : null,
+    rawData: {
+      id,
```
tomkit, line 234: nit: can you add a comment about where this list of fields came from?

PR #785, opened 2023-05-03
feat: add force sync feature and endpoint

tomkit:
- POST to `/mgmt/v1/force-sync?customer_id={}&provide_name={}` to schedule a forced full refresh upon the next sync cycle
- adds openapi spec and endpoints
- migration to add a `resync` flag that is consumable by sync-workers upon the start of the forced sync. consumes the flag upon start: if there's a failure, the user has to reschedule
- adds a lightweight class to help with sync state transitions

## Test Plan

- tested a forced full sync locally against a sync that moved onto the incremental phase of the "full then incremental" strategy and observed that deleted leads re-appeared after the forced sync

## Deployment instructions

### DIFF
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -309,7 +309,7 @@ model CrmEvent {
 model Sync {
   id              String        @id @default(uuid())
   state           Json
-  resync          Boolean       @default(false)
+  forceSyncFlag   Boolean       @default(false)
   strategy        Json
   connectionId    String        @unique @map("connection_id")
   connection      Connection    @relation(fields: [connectionId], references: [id])
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -309,6 +309,7 @@ model CrmEvent {
 model Sync {
   id              String        @id @default(uuid())
   state           Json
+  resync          Boolean       @default(false)
   strategy        Json
   connectionId    String        @unique @map("connection_id")
   connection      Connection    @relation(fields: [connectionId], references: [id])
--- a/packages/sync-workflows/workflows/run_sync.ts
+++ b/packages/sync-workflows/workflows/run_sync.ts
@@ -1,6 +1,6 @@
 import { CommonModel } from '@supaglue/types/common';
 import { CRM_COMMON_MODELS } from '@supaglue/types/crm';
-import { FullThenIncrementalSync, ReverseThenForwardSync, Sync } from '@supaglue/types/sync';
+import { FullThenIncrementalSync, ReverseThenForwardSync } from '@supaglue/types/sync';
 import { ActivityFailure, ApplicationFailure, proxyActivities, uuid4 } from '@temporalio/workflow';
 // Only import the activity types
 import { ImportRecordsResult } from '../activities/import_records';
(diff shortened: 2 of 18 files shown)


packages/sync-workflows/workflows/run_sync.ts
```diff
@@ -63,6 +65,12 @@ export async function runSync({ syncId, connectionId }: RunSyncArgs): Promise<vo
   // Read sync from DB
   const { sync } = await getSync({ syncId });
 
+  if (SyncStateFSM.isResetFlagSet(sync)) {
+    // NOTE: keep the resetted sync state in memory only and let the subsequent sync implementations transition it to the next state
+    SyncStateFSM.setInitialState(sync);
+    await setForceSyncFlag({ syncId }, false);
```
asdfryan, line 71: IMO we should only do this once the whole workflow is completed.

tomkit: Yea I was thinking about it: I think the impact of consuming it at the start is that you can reschedule another forced sync while one is running, but regardless b/c of Schedules, it wouldn't be executed until the next scheduled

So I don't see a strong reason to consume at start or end of a sync. Consuming at the start is easier to grok b/c the call site is all here whereas at the end, we have several termination call sites.

packages/db/prisma/schema.prisma
```diff
@@ -309,6 +309,7 @@ model CrmEvent {
 model Sync {
   id              String        @id @default(uuid())
   state           Json
+  resync          Boolean       @default(false)
```
asdfryan, line 312: Rename to forceSync or forceRefresh or something that matches the endpoint

packages/types/sync.ts
```diff
@@ -78,3 +79,10 @@ export type ReverseThenForwardSyncState =
   | ReverseThenForwardSyncStateForwardPhase;
 
 export type SyncState = FullThenIncrementalSyncState | ReverseThenForwardSyncState;
+
+// The triplet of customer-provided identifiers to uniquely identify a sync. @todo: this triple can also be used elsewhere to uniquel identifer Supaglue resources.
```
asdfryan, line 83: Nit: typo

packages/types/sync.ts
```diff
@@ -3,6 +3,7 @@ import { CommonModel } from './common';
 type BaseSync = {
   id: string;
   connectionId: string;
+  resync: boolean; // flag: whether to treat the sync as phase "created"
```
asdfryan, line 6: rename

packages/sync-workflows/workflows/run_sync.ts
```diff
@@ -323,3 +333,16 @@ const getErrorMessageStack = (err: Error): { message: string; stack: string } =>
...
 };
+
+class SyncStateFSM {
+  static isResetFlagSet(sync: Sync): boolean {
+    return sync.resync;
+  }
+
+  static setInitialState(sync: Sync): Sync {
+    sync.state = {
+      phase: 'created',
```
asdfryan, line 344: Why do we need to do this?

tomkit: There's a state machine that `runSync` goes through. This causes syncs that are in the incremental phase to move back to `created`->`full`

tomkit: We should encapsulate the current direct state manipulations into `SyncStateFSM`

asdfryan: Reading the code more, I don't think the correct thing is to modify the state machine (to be incorrect) but to just `return await doFullStage()` if `sync.forceFull` is set.

Also this will also allow you to get rid of this class which I think is unnecessary extra overhead

tomkit: Incorporated this change

`doFullStage` will still transition the state itself

PR #792, opened 2023-05-04
feat: salesforce CDC

lucasmarshall:


docker-compose.yml
```diff
@@ -28,7 +28,6 @@ services:
       - '5432:5432'
     volumes:
       - pgdata:/var/lib/postgresql/data
-      - ./docker/postgres/docker-entrypoint-initdb.d:/docker-entrypoint-initdb.d
```
lucasmarshall, line 31: Unused

packages/core/lib/logger.ts
```diff
@@ -12,7 +12,7 @@ const addLogContext: (key: string, value: unknown) => void = addContext;
 
 const sentryEnabled = !(process.env.SUPAGLUE_DISABLE_ERROR_REPORTING || process.env.CI);
 
-const LOG_LEVEL = (process.env.SUPAGLUE_LOG_LEVEL ?? 'info') as Level;
+const LOG_LEVEL = process.env.SUPAGLUE_LOG_LEVEL as Level;
```
lucasmarshall, line 15: pino defaults to 'info', so no need to set it here.

packages/core/lib/logger.ts
```diff
@@ -39,5 +39,4 @@ if (sentryEnabled) {
 }
 
 // add a type for this since pino-context doesn't have any
-export const logger: Logger = wrapLogger(pino({}, pino.multistream(streams)));
-logger.level = LOG_LEVEL;
+export const logger: Logger = wrapLogger(pino({ level: LOG_LEVEL }, pino.multistream(streams)));
```
lucasmarshall, line 42: pino still won't log debug logs, even when `level` is set to `'debug'`. Need to debug this further.

.husky/pre-commit
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn turbo run lint-staged "--filter=[HEAD^1]" --output-logs=errors-only
-yarn turbo generate --output-logs=errors-only
+yarn turbo run lint-staged "--filter=[HEAD^1]"
+yarn turbo generate "--filter=!salesforce-pub-sub"
```
lucasmarshall, line 5: Skipping pub-sub here because the proto definitions change very infrequently and we should run generate manually for it

apps/api/routes/internal/v1/index.ts
```diff
@@ -43,7 +43,7 @@ WHERE last_modified_at IS NULL;
       },
     });
     const unsafeSalesforceConnections = await Promise.all(
-      salesforceConnectionModels.map(fromConnectionModelToConnectionUnsafe)
+      salesforceConnectionModels.map(fromConnectionModelToConnectionUnsafe<'salesforce'>)
```
asdfryan, line 46: We should just delete this endpoint IMO. This was for a migration.

apps/salesforce-pub-sub/client.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    logger.debug('starting subscription');
+    const stream = this.grpcClient.subscribe(requestStream(), {
+      headers: this.getHeaders(),
+      timeoutMs: SUBSCRIBE_TIMEOUT_MS,
+    });
+
+    for await (const response of stream) {
+      logger.debug(response, 'got response');
```
asdfryan, line 70: remove?

apps/salesforce-pub-sub/client.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  async *subscribe(fetchRequest: FetchRequestWithRequiredFields) {
+    const requestQueue = [fetchRequest];
+
+    async function* requestStream() {
+      for (;;) {
+        const request = requestQueue.shift();
+        if (request) {
+          logger.debug(request, 'sending request');
```
asdfryan, line 56: remove?

apps/salesforce-pub-sub/client.ts
```diff
@@ -0,0 +1,137 @@
...
+          yield request;
+        }
+        await new Promise((resolve) => setTimeout(resolve, 500));
+      }
+    }
+
+    logger.debug('starting subscription');
```
asdfryan, line 63: remove here and elsewhere

apps/salesforce-pub-sub/index.ts
```diff
@@ -0,0 +1,125 @@
...
+        clientId: integration.config.oauth.credentials.oauthClientId,
+        clientSecret: integration.config.oauth.credentials.oauthClientSecret,
+      }),
+      instanceUrl,
+      refreshToken,
+      maxRequest: 10,
+    });
+
+    // make sure we have the latest accessToken
+    const { access_token: accessToken } = await conn.oauth2.refreshToken(refreshToken);
```
asdfryan, line 54: Do we need to periodically refresh this token?

lucasmarshall: Good catch.

apps/salesforce-pub-sub/index.ts
```diff
@@ -0,0 +1,121 @@
...
+      instanceUrl,
+      refreshToken,
+      maxRequest: 10,
+    });
+
+    let replayId: Uint8Array | undefined;
+
+    const processStream = async (eventType: string, replayPreset = ReplayPreset.LATEST, replayId?: Uint8Array) => {
+      // make sure we have the latest accessToken
+      const { access_token: accessToken } = await conn.oauth2.refreshToken(refreshToken);
```
asdfryan, line 56: I noticed we're refreshing the token here without persisting it in the db (i.e. via emitting the `token_refreshed` event, like we do in the salesforce client.

Does refreshing the token invalidate the existing access token? (e.g. for our syncs etc)

lucasmarshall: Not AFAIK

tomkit (approved):
reviewed offline

PR #795, opened 2023-05-04
chore: refactor services and dependency containers to support multiple verticals

asdfryan:


apps/sync-worker/dependency_container.ts
```diff
@@ -20,36 +18,30 @@ type DependencyContainer = {
...
-  userService: UserService;
-  eventService: EventService;
-  opportunityService: OpportunityService;
   syncService: SyncService;
   syncHistoryService: SyncHistoryService;
   integrationService: IntegrationService;
   applicationService: ApplicationService;
+  crm: {
+    contactService: CrmContactService;
+    accountService: AccountService;
```
tomkit, line 28: nit: if we prefix all of these with `Crm` it will be more explicit for readability before we have engagement services for all of these objects

packages/db/index.ts
```diff
@@ -15,10 +15,15 @@ if (schema) {
...
 
 export const COMMON_MODEL_DB_TABLES = {
-  contacts: `${schemaPrefix}crm_contacts`,
-  accounts: `${schemaPrefix}crm_accounts`,
-  leads: `${schemaPrefix}crm_leads`,
-  opportunities: `${schemaPrefix}crm_opportunities`,
-  users: `${schemaPrefix}crm_users`,
-  events: `${schemaPrefix}crm_events`,
+  crm: {
+    contacts: `${schemaPrefix}crm_contacts`,
```
tomkit, line 19: nit: This is namespacing using the table name -- if we move them into schemas we won't have to do string concatenation

tomkit (approved):

packages/db/index.ts
asdfryan: We don't have separate schemas per vertical at the moment, so we have to do this.

PR #798, opened 2023-05-08
feat: integrate svix for CDC webhooks

lucasmarshall:
Still to do in future PRs:
* Build and deploy new service
* Add subscription endpoints
* Mechanism for picking up new connections without having to restart worker


packages/core/services/index.ts
```diff
@@ -1,7 +1,9 @@
+export * from './application_service';
 export * from './common_models';
 export * from './connection_service';
 export * from './customer_service';
 export * from './integration_service';
 export * from './remote_service';
 export * from './sg_user_service';
 export * from './sync_history_service';
+export * from './webhook_service';
```
tomkit, line 9: Should we also move the existing webhooks into this? Is it meant to encapsulate both? If not, can we name it to reflect it or add a comment

packages/db/prisma/schema.prisma
```diff
@@ -316,3 +317,15 @@ model SyncHistory {
 
   @@map("sync_history")
 }
+
+model ReplayId {
```
tomkit, line 321: Should this be a top-level model? Or be encapsulated in a webhook model?

tomkit (approved):

packages/core/services/index.ts
lucasmarshall: We will want to do that, yes. I'll add a comment

packages/db/prisma/schema.prisma
lucasmarshall: I think this works as top level since the actual webhooks are handled by svix. This is also salesforce specific, so maybe I'll rename

PR #806, opened 2023-05-09
chore: disable analytics in dev

lucasmarshall:


george-xing (approved):

docker-compose.override.dev.yml
```diff
@@ -6,6 +6,7 @@ x-common-env:
   SUPAGLUE_LOG_LEVEL: debug
   SUPAGLUE_DISABLE_ERROR_REPORTING: 1
   NEXT_PUBLIC_SUPAGLUE_DISABLE_ERROR_REPORTING: 1
+  SUPAGLUE_DISABLE_ANALYTICS: 1
```
tomkit, line 9: are these still presence of flags?

lucasmarshall: Yes

tomkit: should it just be `SUPAGLUE_DISABLE_ANALYTICS:` then ?

PR #825, opened 2023-05-12
feat: add remote_id as a _search filter

tomkit:
## Test Plan

[Describe test plan here]

## Deployment instructions


openapi/crm/paths/accounts@_search.yaml
```diff
@@ -21,6 +21,8 @@ post:
               properties:
                 website:
                   $ref: ../components/schemas/filters.yaml#/filter
+                remote_id:
```
asdfryan, line 24: Should probably document that this is an AND and not an OR

asdfryan (approved):

PR #834, opened 2023-05-15
feat: create connection by passing in creds manually

albertyfwu:
We already have this in `internal`, but we need to expose this in `mgmt` for use in Cloud temporarily for migrators from self-hosted to Cloud.

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/openapi/mgmt/paths/connections.yaml
+++ b/openapi/mgmt/paths/connections.yaml
@@ -37,6 +37,28 @@ get:
                   category: crm
                   remote_id: 1234567
                   instance_url: https://app.hubspot.com/contacts/1234567
+post:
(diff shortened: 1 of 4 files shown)


openapi/mgmt/paths/connections.yaml
```diff
@@ -37,6 +37,28 @@ get:
...
+  tags:
+    - Connections
+  parameters: []
+  security:
+    - ApiKeyAuth: []
+  requestBody:
+    required: true
+    content:
+      application/json:
+        schema:
```
albertyfwu, line 52: not defining very strict types since we will be removing this endpoint soon

PR #837, opened 2023-05-15
fix: pub-sub worker issues on cloud

lucasmarshall:


apps/salesforce-pub-sub/index.ts
lucasmarshall, line 1: Adding awaits fixes error logging stack traces

albertyfwu (approved):

apps/salesforce-pub-sub/index.ts
```diff
@@ -102,19 +103,35 @@ const { connectionService, integrationService, webhookService, applicationServic
       }
     };
 
-    const subscribe = (eventType: string) => {
-      const streamErrorHandler = (err: any) => {
+    const subscribe = async (eventType: string) => {
```
tomkit, line 106: What's diff between local and cloud for this?

tomkit (approved):

apps/salesforce-pub-sub/index.ts
lucasmarshall: See my comment above

lucasmarshall: Making this async fixes some stack trace logging issues

lucasmarshall: The error handling were errors I only saw on cloud (we have way more varied connections on cloud that in local testing)

PR #842, opened 2023-05-16
fix: add more context to logs, and add context when we have it - not later

lucasmarshall:
Also update turbo

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,4 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 yarn turbo run lint-staged "--filter=[HEAD^1]"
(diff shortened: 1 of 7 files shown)


.husky/pre-commit
```diff
@@ -2,4 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 yarn turbo run lint-staged "--filter=[HEAD^1]"
-yarn turbo generate "--filter=!salesforce-pub-sub"
+yarn turbo run generate "--filter=!salesforce-pub-sub"
```
lucasmarshall, line 5: Turbo now has a `generate` subcommand of it's own. Use `run` to disambiguate.

albertyfwu (approved):

asdfryan (approved):

PR #843, opened 2023-05-16
feat: support POST and PATCH for pipedrive

asdfryan:
Testing locally


packages/core/remotes/crm/pipedrive/index.ts
```diff
@@ -248,14 +271,141 @@ class PipedriveClient extends AbstractCrmRemoteClient {
     commonModelType: T,
     params: CRMCommonModelTypeMap<T>['createParams']
   ): Promise<CRMCommonModelTypeMap<T>['object']> {
-    throw new Error('Not implemented');
+    switch (commonModelType) {
+      case 'contact':
+        return this.createContact(params as RemoteContactCreateParams);
```
albertyfwu, line 276: why do you need to assert the type here? we don't need to do this for hubspot or salesforce

packages/core/remotes/crm/pipedrive/mappers.ts
```diff
@@ -139,17 +144,126 @@ export const fromPipedrivePhonesToPhoneNumbers = (phoneNumbers: { label: string;
...
+    !organization.address_locality &&
+    !organization.address_admin_area_level_1 &&
+    !organization.address_postal_code &&
+    !organization.address_country
+  ) {
+    return [];
+  }
+  return [
+    {
+      street1: `${organization.address_street_number ?? ''} ${organization.address_route ?? ''}`,
```
albertyfwu, line 161: if one of these is null, there will be an extra space. is that OK?

albertyfwu (approved):

PR #855, opened 2023-05-17
fix: heartbeat on salesforce poll

asdfryan:


tomkit (approved):

packages/core/remotes/crm/base.ts
```diff
@@ -26,7 +26,11 @@ export abstract class AbstractCrmRemoteClient extends AbstractRemoteClient imple
     return 'crm';
   }
 
-  abstract listObjects(commonModelType: CRMCommonModelType, updatedAfter?: Date): Promise<Readable>;
+  abstract listObjects(
+    commonModelType: CRMCommonModelType,
+    updatedAfter?: Date,
+    onPoll?: () => void
```
albertyfwu, line 32: maybe we should just name this `heartbeat` and break the abstraction barrier, so the method is expected to just heartbeat as often as it can when it's making forward progress

albertyfwu (approved):

lucasmarshall (approved):

PR #871, opened 2023-05-18
chore: code to write common models to postgres

albertyfwu:
this is code to write common models to postgres; it's not used yet.

next PR will be a new temporal workflow/schedule that will periodically sync data to postgres destination

## Test Plan

[Describe test plan here]

## Deployment instructions


openapi/crm/components/schemas/objects/addresses.yaml
```diff
@@ -26,11 +26,11 @@ items:
       type: string
       nullable: true
       example: CA
-    street1:
+    street_1:
```
asdfryan, line 29: I think this change is not backwards compatible -- in order to migrate we should probably keep both around / talk with our customers etc.

PR #886, opened 2023-05-19
chore: destination_id on integration in prisma.schema

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -26,13 +26,14 @@ model Application {
 }
 
 model Destination {
-  id            String      @id @default(uuid())
+  id            String        @id @default(uuid())
   type          String
-  applicationId String      @unique @map("application_id")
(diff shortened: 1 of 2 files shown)


packages/db/prisma/schema.prisma
```diff
@@ -26,13 +26,14 @@ model Application {
 }
 
 model Destination {
-  id            String      @id @default(uuid())
+  id            String        @id @default(uuid())
   type          String
-  applicationId String      @unique @map("application_id")
-  application   Application @relation(fields: [applicationId], references: [id], onDelete: Cascade)
+  applicationId String        @unique @map("application_id")
```
asdfryan, line 31: should we drop this? or are we keeping it for now.

albertyfwu: going to drop this in a future PR. i already ran this migration locally and would prefer not to reset my DB / undo migration / etc.

asdfryan (approved):

PR #927, opened 2023-05-24
chore: internal endpoint for sync migration

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/api/routes/internal/v1/index.ts
+++ b/apps/api/routes/internal/v1/index.ts
@@ -1,3 +1,4 @@
+import { getDependencyContainer } from '@/dependency_container';
 import { internalMiddleware } from '@/middleware/internal';
 import { internalApplicationMiddleware } from '@/middleware/internal_application';
 import { orgHeaderMiddleware } from '@/middleware/org';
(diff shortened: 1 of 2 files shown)


apps/api/routes/internal/v1/index.ts
```diff
@@ -12,13 +13,97 @@ import syncHistory from './sync_history';
...
+      },
+    });
+
+    await prisma.$transaction([
+      prisma.sync.updateMany({
+        data: {
+          // TODO: we need to kill the old syncs first before we set this,
+          // since it could be overridden by the old running syncs
+          strategy: {
+            type: 'full then incremental',
```
asdfryan, line 84: We need to condition this on whether or not the destination is s3 (`full only`)

apps/api/routes/mgmt/v1/sync_migration.ts
```diff
@@ -61,6 +61,14 @@ export default function init(app: Router) {
           id: existingSync.id,
         },
         data: {
+          // TODO: we need to kill the old syncs first before we set this,
+          // since it could be overridden by the old running syncs
+          strategy: {
+            type: 'full then incremental',
```
asdfryan, line 67: same here

asdfryan (approved):

PR #928, opened 2023-05-24
feat: support full only for S3 destinations

asdfryan:


packages/sync-workflows/workflows/run_sync.ts
```diff
@@ -125,6 +129,54 @@ export async function runSync({ syncId, connectionId, category }: RunSyncArgs):
   );
 }
 
+async function doFullOnlySync({
```
albertyfwu, line 132: you need to implement this in `run_managed_sync` not `run_sync`

PR #930, opened 2023-05-25
chore: kill temporal schedules before migrating connections

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


apps/api/services/connection_and_sync_service.ts
```diff
@@ -262,6 +270,86 @@ export class ConnectionAndSyncService {
     }
   }
 
+  // TODO: Remove when we're done migrating
+  async deleteTemporalSyncsV1(syncIds: string[]): Promise<void> {
```
albertyfwu, line 274: this code is copied from `sync_service` in sync-worker

asdfryan (approved):

PR #931, opened 2023-05-25
fix: backfill correct strategy

asdfryan:


apps/api/routes/mgmt/v1/sync_migration.ts
```diff
@@ -44,6 +45,25 @@ export default function init(app: Router) {
       });
     }
 
+    let strategyType = 'full then incremental';
+    if (version === 'v2') {
+      const { destination } = await prisma.integration.findFirstOrThrow({
```
albertyfwu, line 50: why not just query
```
prisma.connection.findFirstOrThrow({
  select: {
    integration: {
      select: {
        destination: true
      }
    }
  }
});
```

albertyfwu (approved):

PR #946, opened 2023-05-31
feat: custom objects and schemas

albertyfwu:
issues:
- need to be part of pilot program to create custom object schemas: https://developers.hubspot.com/docs/api/custom-objects-schema-pilot
- not able to update properties yet. need to implement that
- when fetching custom object, it doesn't return the fields that we set, since hubspot expects us to pass in the properties we want explicitly

## Test Plan

[Describe test plan here]

## Deployment instructions


tomkit (approved):
🙏

PR #950, opened 2023-06-02
feat(mgmt): add pagination to sync history

tomkit:
- server-side pagination for sync history mgmt ui page with 100 records at a time
- make the datagrid compact
- disallow client-side  sorting/filtering which will cause trouble when mixed with server pagination

https://github.com/supaglue-labs/supaglue/assets/471516/63f75af7-d9a2-457e-823f-e40664bea334

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/mgmt-ui/src/pages/applications/[applicationId]/sync_logs/index.tsx
+++ b/apps/mgmt-ui/src/pages/applications/[applicationId]/sync_logs/index.tsx
@@ -1,5 +1,4 @@
 import LogsTable from '@/components/logs/LogsTable';
-import Spinner from '@/components/Spinner';
 import { useSyncHistory } from '@/hooks/useSyncHistory';
 import Header from '@/layout/Header';
 import { getServerSideProps } from '@/pages/applications/[applicationId]';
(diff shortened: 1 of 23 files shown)


packages/types/common.ts
```diff
@@ -44,6 +44,7 @@ export type PaginatedResult<T> = {
   next: string | null;
   previous: string | null;
   results: T[];
+  count: number;
```
asdfryan, line 47: It's unclear from this name whether this is referring to the total count, or the count in just this result.

apps/mgmt-ui/src/pages/applications/[applicationId]/sync_logs/index.tsx
```diff
@@ -29,7 +41,13 @@ export default function Home() {
       <Box sx={{ flex: 1, display: 'flex', flexDirection: 'column' }}>
         <Header title="Sync Logs" onDrawerToggle={handleDrawerToggle} />
         <Box component="main" sx={{ flex: 1, py: 6, px: 4, bgcolor: '#eaeff1' }}>
-          {isLoading ? <Spinner /> : <LogsTable data={syncHistories?.results ?? []} />}
+          <LogsTable
+            handleNextPage={handleNextPage}
+            handlePreviousPage={handlePreviousPage}
+            rowCount={syncHistories?.totalCount || 0}
```
Comment (albertyfwu) [inline apps/mgmt-ui/src/pages/applications/[applicationId]/sync_logs/index.tsx:47]: nit: `??` instead

packages/core/services/sync_history_service.ts
```diff
@@ -127,7 +127,7 @@ export class SyncHistoryService {
     const connections = await this.#connectionService.listSafe(applicationId, customerId, providerName);
```
albertyfwu, line 127: maybe beyond scope of this PR, but it seems like we don't need to make this separate call to get connections and just query for it in the prisma call

albertyfwu (approved):

PR #957, opened 2023-06-05
feat: actions api for engagement vertical

asdfryan:


apps/api/routes/crm/v2/user.ts
```diff
@@ -23,5 +24,13 @@ export default function init(app: Router): void {
     }
   );
 
+  router.post('/', async (req: Request, res: Response) => {
+    throw new BadRequestError('POST not supported for /users');
+  });
+
+  router.patch('/:mailbox_id', async (req: Request, res: Response) => {
+    throw new BadRequestError('PATCH not supported for /users');
```
albertyfwu, line 32: nit: do you need to put `/users` in the error message? seems evident from the request URL itself

asdfryan: I think it's helpful since it's supported for all the other models

PR #963, opened 2023-06-06
fix: some things I noticed while debugging the mgmt-ui image

lucasmarshall:


.dockerignore
```diff
@@ -2,3 +2,4 @@
 .yarn/cache
 **/node_modules
 docs
+**/.next
```
lucasmarshall, line 5: Makes the build context way smaller and speeds up docker builds

docs/docs/getting-started.mdx
```diff
@@ -44,6 +44,9 @@ docker compose up
 
 Once Supaglue is running locally, open the [management portal](http://localhost:3000) in your browser.
 
+- username: `admin`
+- password: `admin`
```
lucasmarshall, line 48: We didn't tell them how to login to the interface when running locally.

asdfryan (approved):

PR #970, opened 2023-06-06
chore(mgmt-ui): add number of synced records to sync log page

tomkit:
- `num_records_synced` for each sync, `null` is for start of sync and remains such if it fails, otherwise the number of records is used (same value used for the webhook payload)

Screenshot 2023-06-06 at 3 00 44 PM

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/openapi/v2/mgmt/components/schemas/sync_history.yaml
+++ b/openapi/v2/mgmt/components/schemas/sync_history.yaml
@@ -34,6 +34,10 @@ properties:
       - SUCCESS
       - IN_PROGRESS
       - FAILURE
+  num_records_synced:
+    type: number
+    nullable: true
+    example: 100
 required:
   - model_name
   - start_timestamp
--- a/packages/sync-workflows/workflows/run_managed_sync.ts
+++ b/packages/sync-workflows/workflows/run_managed_sync.ts
@@ -98,6 +98,7 @@ export async function runManagedSync({ syncId, connectionId, category }: RunMana
           status: 'FAILURE',
           errorMessage,
           errorStack,
+          numRecordsSynced: null,
         });
         await maybeSendSyncFinishWebhook({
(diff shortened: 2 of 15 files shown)


openapi/v2/mgmt/components/schemas/sync_history.yaml
```diff
@@ -34,6 +34,9 @@ properties:
       - SUCCESS
       - IN_PROGRESS
       - FAILURE
+  num_records_synced:
```
asdfryan, line 37: Does this need to be nullable? required?

packages/sync-workflows/workflows/run_managed_sync.ts
```diff
@@ -120,14 +121,22 @@ export async function runManagedSync({ syncId, connectionId, category }: RunMana
 
   await Promise.all(
     getCommonModels(category).map(async (commonModel) => {
-      await logSyncFinish({ syncId, connectionId, historyId: historyIdsMap[commonModel], status: 'SUCCESS' });
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore `numRecordsSyncedMap` is indeed defined here
```
asdfryan, line 125: What is the ts error?

asdfryan (changes requested):
Looks like you need to fix your openapi spec

openapi/v2/mgmt/components/schemas/sync_history.yaml
tomkit: yes, nullable, good catch

packages/sync-workflows/workflows/run_managed_sync.ts
tomkit: this was an existing ignore that i propagated

asdfryan (approved):

PR #971, opened 2023-06-06
chore: use new temporal instead of temporalite

albertyfwu:
temporalite is being deprecated

ideally, we should just `brew install temporal` and use that, but that makes it harder to start up, since we can't just run `docker compose up` and need to run another script too.

Thoughts?

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/docker-compose.override.dev.yml
+++ b/docker-compose.override.dev.yml
@@ -48,4 +48,4 @@ services:
     volumes:
       - .:/app
     working_dir: /app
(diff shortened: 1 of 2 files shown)


albertyfwu:
@lucasmarshall is there a better way of doing this?

lucasmarshall (approved):
I had a branch where I was doing this exact thing, but ran into some issues. This LGTM

PR #973, opened 2023-06-07
chore(mgmt-ui): remove internal route versioning

tomkit:
- internal routes are not a public interface, let's make them it version-less

## Test Plan

[Describe test plan here]

## Deployment instructions


albertyfwu (approved):
thanks

lucasmarshall (approved):

PR #977, opened 2023-06-07
chore: provider type and mappers

asdfryan:


packages/types/oauth.ts
```diff
@@ -0,0 +1,14 @@
+export type OauthConfigDecrypted = {
+  oauthScopes: string[];
+  credentials: OauthCredentials;
+};
+
+export type OauthConfigEncrypted = {
+  oauthScopes: string[];
+  credentials: string;
```
albertyfwu, line 8: is it a Buffer?

packages/types/provider.ts
```diff
@@ -0,0 +1,42 @@
+import { OauthConfigDecrypted, OauthConfigEncrypted } from '.';
```
albertyfwu, line 1: `Oauth` -> `OAuth` everywhere?

albertyfwu (approved):

packages/types/oauth.ts
asdfryan: It's not at the moment. Migration could be a pain -- would rather keep it string.

PR #979, opened 2023-06-08
feat(connectors): add ms dynamics 365 sales

tomkit:
- add back dynamics integration card
- fix some oauth authorize scope issues with azure AD
Screenshot 2023-06-07 at 4 39 10 PM
Screenshot 2023-06-07 at 4 39 25 PM

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/api/routes/oauth.ts
+++ b/apps/api/routes/oauth.ts
@@ -1,7 +1,7 @@
 import { getDependencyContainer } from '@/dependency_container';
 import { Client as HubspotClient } from '@hubspot/api-client';
 import { BadRequestError } from '@supaglue/core/errors';
-import { getConnectorAuthConfig } from '@supaglue/core/remotes';
+import { getAdditionalConnectorAuthConfig, getConnectorAuthConfig } from '@supaglue/core/remotes';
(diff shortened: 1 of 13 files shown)


apps/api/routes/oauth.ts
```diff
@@ -153,6 +153,7 @@ export default function init(app: Router): void {
       const { oauthClientId, oauthClientSecret } = integration.config.oauth.credentials;
 
       const auth = getConnectorAuthConfig(integration.category, providerName);
+      const additionalAuth = getAdditionalConnectorAuthConfig(integration.category, providerName);
```
tomkit, line 156: `auth` is used directly for `simple-oauth2` and it doesn't allow extraneous fields so placing other configurations into another object

apps/api/routes/oauth.ts
```diff
@@ -76,13 +76,13 @@ export default function init(app: Router): void {
 
       const authorizationUri = client.authorizeURL({
         redirect_uri: REDIRECT_URI,
-        scope: oauthScopes.join(' '),
```
tomkit, line 79: oauthScopes is already an array

PR #984, opened 2023-06-08
chore: partial work for syncing records from hubspot

albertyfwu:
Still need to:
- Pull associations for objects
- Pull out some fields that will be necessary to write into postgres not into the JSONB column (need to define type/interface to pull out id, last_modified_at, etc.)

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/lib/ratelimit.ts
+++ b/packages/core/lib/ratelimit.ts
@@ -4,7 +4,7 @@ import { ASYNC_RETRY_OPTIONS, logger } from '.';
 import { TooManyRequestsError } from '../errors';
 
 export const isAxiosRateLimited = (e: any): boolean => {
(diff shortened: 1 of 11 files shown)


packages/core/lib/ratelimit.ts
```diff
@@ -4,7 +4,7 @@ import { ASYNC_RETRY_OPTIONS, logger } from '.';
 import { TooManyRequestsError } from '../errors';
 
 export const isAxiosRateLimited = (e: any): boolean => {
-  return isAxiosError(e) && e.status === 429;
+  return isAxiosError(e) && e.response?.status === 429;
```
albertyfwu, line 7: fix

asdfryan (approved):

PR #985, opened 2023-06-08
chore: sync config openapi, endpoints, services, types, backfill

asdfryan:
tested locally


apps/api/routes/mgmt/v2/sync_config/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      const syncConfig = await syncConfigService.getByIdAndApplicationId(
+        req.params.sync_config_id,
+        req.supaglueApplication.id
+      );
+      return res.status(200).send(snakecaseKeys(syncConfig));
+    }
+  );
+
+  syncConfigRouter.put(
+    '/:sync_config_id',
```
albertyfwu, line 68: did we want to use global sync config? or keep it this way for now?

albertyfwu (approved):

apps/api/routes/mgmt/v2/sync_config/index.ts
asdfryan: But it's not at all global? It only applies for a single provider.

PR #988, opened 2023-06-12
feat(connectors): add ms dynamics 365 sales oauth 2

tomkit:
Screenshot 2023-06-07 at 4 39 10 PM
Screenshot 2023-06-07 at 4 39 25 PM

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/api/routes/oauth.ts
+++ b/apps/api/routes/oauth.ts
@@ -1,7 +1,7 @@
 import { getDependencyContainer } from '@/dependency_container';
 import { Client as HubspotClient } from '@hubspot/api-client';
 import { BadRequestError } from '@supaglue/core/errors';
-import { getAdditionalConnectorAuthConfig, getConnectorAuthConfig } from '@supaglue/core/remotes';
+import { getConnectorAuthConfig } from '@supaglue/core/remotes';
 import { ConnectionCreateParamsAny, ConnectionUpsertParamsAny, ProviderName } from '@supaglue/types';
 import { CRMProviderName, SUPPORTED_CRM_CONNECTIONS } from '@supaglue/types/crm';
 import { EngagementProviderName, SUPPORTED_ENGAGEMENT_CONNECTIONS } from '@supaglue/types/engagement';
@@ -161,7 +161,6 @@ export default function init(app: Router): void {
       const { oauthClientId, oauthClientSecret } = integration.config.oauth.credentials;
 
       const auth = getConnectorAuthConfig(integration.category, providerName);
-      const additionalAuth = getAdditionalConnectorAuthConfig(integration.category, providerName);
 
       if (loginUrl) {
         auth.tokenHost = loginUrl;
@@ -173,12 +172,14 @@ export default function init(app: Router): void {
         auth.authorizePath = `${auth.authorizePath}?${loginParams}`;
       }
 
+      const { authorizeWithScope, ...simpleOauth2Auth } = auth;
+
       const client = new simpleOauth2.AuthorizationCode({
         client: {
           id: oauthClientId,
           secret: oauthClientSecret,
         },
-        auth,
+        auth: simpleOauth2Auth,
         options: {
           authorizationMethod: 'body' as AuthorizationMethod,
         },
@@ -190,7 +191,7 @@ export default function init(app: Router): void {
       const tokenWrapper = await client.getToken({
         code,
         redirect_uri: REDIRECT_URI,
-        scope: additionalAuth.authorizeWithScope ? scope : undefined,
+        scope: auth.authorizeWithScope ? scope : undefined,
         ...additionalAuthParams,
       });
 
--- a/apps/api/routes/oauth.ts
+++ b/apps/api/routes/oauth.ts
@@ -1,7 +1,7 @@
 import { getDependencyContainer } from '@/dependency_container';
 import { Client as HubspotClient } from '@hubspot/api-client';
 import { BadRequestError } from '@supaglue/core/errors';
-import { getConnectorAuthConfig } from '@supaglue/core/remotes';
+import { getAdditionalConnectorAuthConfig, getConnectorAuthConfig } from '@supaglue/core/remotes';
 import { ConnectionCreateParamsAny, ConnectionUpsertParamsAny, ProviderName } from '@supaglue/types';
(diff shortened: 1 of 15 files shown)


apps/api/routes/oauth.ts
```diff
@@ -76,14 +81,15 @@ export default function init(app: Router): void {
 
       const authorizationUri = client.authorizeURL({
         redirect_uri: REDIRECT_URI,
-        scope: oauthScopes.join(' '),
```
tomkit, line 79: oauthScopes is already a `string[]`

packages/core/remotes/crm/ms_dynamics_365_sales/index.ts
```diff
@@ -48,7 +48,11 @@ export function newClient(
 
 export const authConfig: ConnectorAuthConfig = {
   tokenHost: 'https://login.microsoftonline.com',
-  tokenPath: '/common/oauth2/v2.0/token',
```
tomkit, line 51: Using v1 oauth2 endpoints to scope the authentication to the resource={dynamics host name}

apps/api/routes/oauth.ts
```diff
@@ -176,6 +190,7 @@ export default function init(app: Router): void {
       const tokenWrapper = await client.getToken({
         code,
         redirect_uri: REDIRECT_URI,
+        scope: additionalAuth.authorizeWithScope ? scope : undefined,
```
tomkit, line 193: Only Dynamics require this. Passing it when it's not required causes errors with Salesforce and potentially others

apps/api/routes/oauth.ts
```diff
@@ -76,14 +81,15 @@ export default function init(app: Router): void {
 
       const authorizationUri = client.authorizeURL({
         redirect_uri: REDIRECT_URI,
-        scope: oauthScopes.join(' '),
+        scope: oauthScopes,
```
asdfryan, line 84: Can you test that this does not break hubspot, salesforce, outreach, and pipedrive oauth flows?

apps/api/routes/oauth.ts
asdfryan: If that's the case, can you condition on the provider so that this code is more readable?

apps/api/routes/oauth.ts
tomkit: I tested this on Salesforce which should extrapolate to the others since the integration scope configuration is the same for all

docs/docs/platform/managed-auth.mdx
```diff
@@ -22,28 +22,29 @@ An embedded link takes the following form:
...
+| `{PROVIDER_NAME}`  | The name of the third-party provider (e.g. `salesforce`, `hubspot`, etc.)                                                                                                                                                                                    | Yes      |
+| `{RETURN_URL}`     | The URL to return to once the OAuth connection is complete.<br/> <b>NOTE:</b> this should be uriEncoded                                                                                                                                                      | Yes      |
+| `{LOGIN_URL}`      | The Oauth2 hostname for authorize and token endpoints. <br/> This can be used for Salesforce sandbox accounts and should be `https://test.salesforce.com` for that use case.<br/> <b>NOTE:</b> this should be uriEncoded                                     | No       |
+| `{LOGIN_PARAMS}`   | The query parameters to be added for Oauth2 authorize and token URLs. <br/> This can be used for MS Dynamics 365 Sales accounts to specify the `resource`, e.g. `resource=https://org8d6f84ed.crm.dynamics.com`.<br/> <b>NOTE:</b> this should be uriEncoded | No       |
```
lucasmarshall, line 47: How is a developer supposed to know what this is for a certain customer? Can't we just use the MS Dynamics discovery service to get this?

lucasmarshall (changes requested):

docs/docs/platform/managed-auth.mdx
lucasmarshall: https://learn.microsoft.com/en-us/previous-versions/dynamicscrm-2016/developers-guide/mt607485(v=crm.8)

packages/types/connection.ts
```diff
@@ -16,6 +16,10 @@ export type ConnectionCredentialsDecrypted<T extends ProviderName> = BaseConnect
         loginUrl?: string;
       }
     : T extends 'pipedrive'
+    ? {
+        instanceUrl: string;
+      }
+    : T extends 'ms_dynamics_365_sales'
```
lucasmarshall, line 22:
```suggestion
    : T extends 'pipedrive' | 'ms_dynamics_365_sales'
```

lucasmarshall (changes requested):

docs/docs/platform/managed-auth.mdx
tomkit: It should be part of their UI to ask for the customer to enter their account hostname

packages/core/remotes/crm/base.ts
```diff
@@ -127,7 +127,12 @@ export type ConnectorAuthConfig = {
   authorizePath: string;
 };
 
+export type AdditionalConnectorAuthConfig = {
+  authorizeWithScope: boolean;
+};
+
 export type CrmConnectorConfig<T extends CRMProviderName> = {
+  additionalAuthConfig: AdditionalConnectorAuthConfig;
```
albertyfwu, line 135: Is this a general bag of "additional auth configs" that are required for any individual connector? if so, how is this different from just sticking it in the `ConnectorAuthConfig`?

If we want to have this only be for dynamics, then shouldn't we put it in a dynamics-only config?

We should either put `authorizeWithScope` as a field required on all ConnectorAuthConfig or only on dynamics. What is the benefit of having this in-between AdditionalConnectorAuthConfig?

packages/types/connection.ts
tomkit: I duplicated it so it's more flexible if we want to add additional fields on a per-provider basis

A few thoughts on the object model here:
- If we just make `instanceUrl` required then we can get rid of generics/types for providers

packages/core/remotes/crm/base.ts
tomkit: It's a bag

The consumer of `ConnectorAuthConfig` is simple-oauth2, which chokes on superfluous fields on the object

lucasmarshall (approved):

PR #991, opened 2023-06-13
feat(mgmt-ui): ability to delete customers and connections. connections

tomkit:
page.

https://github.com/supaglue-labs/supaglue/assets/471516/6cc72d21-c78f-4b0b-a08d-d5e3711877d3

[Describe your change here]

## Test Plan

[Describe test plan here]

## Deployment instructions


apps/mgmt-ui/src/hooks/useActiveCustomerId.ts
```diff
@@ -0,0 +1,10 @@
+import { useRouter } from 'next/router';
+
+export function useActiveCustomerId(): string {
+  const router = useRouter();
```
albertyfwu, line 4: is `router.query` always immediately populated? Is it possible for `customerId` to be `undefined` when the component is initially mounted? I recall that we had issues before because it is temporarily `undefined`i

albertyfwu (approved):

apps/mgmt-ui/src/hooks/useActiveCustomerId.ts
tomkit: it's a hook so it might not be - we don't wait for the entire page to render right now before letting users interact with it, but that should only be problematic with really slow connections

PR #993, opened 2023-06-13
fix(core): remove postgres pool cache

tomkit:
- remove caching for now since pg pool cache isn't invalidated right now. some customers run into this issue

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/destination_writers/postgres.ts
+++ b/packages/core/destination_writers/postgres.ts
@@ -29,8 +29,6 @@ import { logger } from '../lib';
 import { BaseDestinationWriter, WriteCommonModelsResult } from './base';
 import { getSnakecasedKeysMapper } from './util';
 
-const destinationIdToPool: Record<string, Pool> = {};
-
 export class PostgresDestinationWriter extends BaseDestinationWriter {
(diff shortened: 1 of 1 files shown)


packages/core/destination_writers/postgres.ts
```diff
@@ -40,13 +38,7 @@ export class PostgresDestinationWriter extends BaseDestinationWriter {
   }
 
   async #getClient(): Promise<PoolClient> {
-    const existingPool = destinationIdToPool[this.#destination.id];
-    if (existingPool) {
-      return await existingPool.connect();
-    }
-
     const pool = new Pool(this.#destination.config);
```
albertyfwu, line 41: if we remove the cache, we might want to be careful that we don't hold onto connections for too long lest we exhaust our customers' connection limits. are the default settings safe for our usage pattern?

albertyfwu (approved):

packages/core/destination_writers/postgres.ts
tomkit: good Q- haven't verified, let's do that on staging with our load tests

PR #995, opened 2023-06-13
chore: replace all usage of `integration` with `provider` and `syncConfig`

asdfryan:
## Test Plan

Tested locally

## Deployment instructions


apps/api/services/connection_and_sync_service.ts
```diff
@@ -88,35 +82,22 @@ export class ConnectionAndSyncService {
...
-    );
-    let strategyType = 'full then incremental';
-    if (version === 'v2' && integration.destinationId) {
-      const destination = await this.#destinationService.getDestinationById(integration.destinationId);
-      if (destination.type === 's3') {
-        strategyType = 'full only';
-      }
-    }
-    const application = await this.#applicationService.getById(integration.applicationId);
+    const provider = await this.#providerService.getByNameAndApplicationId(params.providerName, params.applicationId);
```
asdfryan, line 98: We no longer need to do this destination check since the strategy is encoded in the sync config.

packages/core/remotes/crm/base.ts
```diff
@@ -69,7 +70,8 @@ export abstract class AbstractCrmRemoteClient extends AbstractRemoteClient imple
   abstract listCommonModelRecords(
     commonModelType: CRMCommonModelType,
     updatedAfter?: Date,
-    heartbeat?: () => void
+    heartbeat?: () => void,
+    fetchAllFields?: boolean
```
asdfryan, line 74: We had to add this field to the function itself since it's no longer part of the provider config.

packages/db/prisma/migrations/20230612220530_connection_updates/migration.sql
```diff
@@ -0,0 +1,21 @@
+/*
```
albertyfwu, line 1: you'll need to re-do these migrations since a new migration has already landed on `main` since you started working on this PR:
`20230613033542_remove_remote_id`

apps/api/routes/internal/index.ts
```diff
@@ -27,7 +27,7 @@ export default function init(app: Router): void {
   v1ApplicationRouter.get('/_migration_connections', async (req, res) => {
     const connections = await prisma.connection.findMany({
       where: {
-        integration: {
+        provider: {
```
albertyfwu, line 30: why do we need to update this to `provider`? Can't we keep this `integration` until customers are done migrating?

apps/api/services/connection_and_sync_service.ts
```diff
@@ -123,38 +107,48 @@ export class ConnectionAndSyncService {
...
-          data: {
-            syncId,
-          },
-        }),
-      ]);
+          })
+        );
+      }
+
+      const [connectionModel] = await this.#prisma.$transaction(txns);
```
albertyfwu, line 151: nit: i think there's an alternative usage of `.$transaction` where you can pass in a callback that takes `transaction` as an arg so you don't have to do this if/else array creation

packages/types/sync.ts
```diff
@@ -4,6 +4,7 @@ import { EngagementCommonModelType } from './engagement';
 type BaseSync = {
   id: string;
   connectionId: string;
+  syncConfigId?: string;
```
albertyfwu, line 7: is it a TODO to make this required later?

packages/sync-workflows/activities/get_destination.ts
```diff
@@ -2,16 +2,16 @@ import { DestinationService } from '@supaglue/core/services/destination_service'
 import { Destination } from '@supaglue/types';
 
 export type GetDestinationArgs = {
-  connectionId: string;
+  syncId: string;
```
albertyfwu, line 5: given that this is backwards-incompatible, you'll need to make sure to kill existing workflows.

otherwise, you could make this backwards-compatible by adding `syncId` as an optional field and then deprecating `connectionId` later

apps/sync-worker/services/sync_service.ts
```diff
@@ -91,23 +92,23 @@ export class SyncService {
   }
 
   public async processSyncChanges(): Promise<void> {
-    // Get all the IntegrationChange objects
-    const integrationChanges = await this.#prisma.integrationChange.findMany();
-    const integrationChangeIds = integrationChanges.map((integrationChange) => integrationChange.id);
+    // Get all the SyncConfigChange objects
+    const syncConfigChanges = await this.#prisma.syncConfigChange.findMany();
```
albertyfwu, line 96: nit: if we're only using the `ids`, maybe we can `select: { id: true }`

apps/api/services/connection_and_sync_service.ts
```diff
@@ -123,38 +107,48 @@ export class ConnectionAndSyncService {
...
+      const txns: Prisma.PrismaPromise<any>[] = [
         this.#prisma.connection.create({
           data: {
             id: connectionId,
             category: params.category,
             providerName: params.providerName,
             customerId: getCustomerIdPk(params.applicationId, params.customerId),
-            integrationId: integration.id,
+            // TODO: Delete
+            integrationId: provider.id,
```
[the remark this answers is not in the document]
asdfryan: It's not but I think it doesn't matter what we write here since we will get rid of it (we don't actually have access to the actual integration ID).

packages/types/sync.ts
asdfryan: Yes

packages/sync-workflows/activities/get_destination.ts
asdfryan: What do you recommend here?

albertyfwu: personally i would make sure to kill existing workflows after deployment

apps/api/services/connection_and_sync_service.ts
```diff
@@ -107,8 +107,8 @@ export class ConnectionAndSyncService {
       const connectionId = uuidv4();
       const syncId = uuidv4();
 
-      const txns: Prisma.PrismaPromise<any>[] = [
-        this.#prisma.connection.create({
+      const connectionModel = await this.#prisma.$transaction(async () => {
```
albertyfwu, line 110: are you sure this works? i think you need to use the `prisma` that's passed in as an arg to your callback fn

albertyfwu: https://www.prisma.io/docs/concepts/components/prisma-client/transactions#interactive-transactions

asdfryan: Nice catch

albertyfwu (approved):

apps/api/services/connection_and_sync_service.ts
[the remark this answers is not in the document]
albertyfwu: why aren't we just deleting this?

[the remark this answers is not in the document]
asdfryan: We still need this for v1 syncs unfortunately.

packages/core/services/sync_config_service.ts
```diff
@@ -47,26 +65,61 @@ export class SyncConfigService {
     return fromSyncConfigModel(syncConfig);
   }
```
albertyfwu, line 67: are we migrating all syncs to have `syncConfigId`? If so, shouldn't this be an error?

asdfryan: Not the v1 syncs because they lack destinations.

packages/core/remotes/crm/base.ts
albertyfwu: why is it no longer on the provider config?

asdfryan: It is part of the `SyncConfig`, which used to be part of the `IntegrationConfig` but is now its own thing. Because the remoteClient only requires a `Provider` object (e.g. for connections that don't have syncs), we need to inject it into the function parameters instead.

PR #996, opened 2023-06-13
chore: associations for hubspot and temporal activity to write raw records

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


packages/core/remotes/crm/salesforce/index.ts
```diff
@@ -293,7 +318,9 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
             return callback(e);
           }
         },
-      })
+      }),
```
albertyfwu, line 321: @tomkit this is the bug

asdfryan (approved):

PR #1003, opened 2023-06-13
fix: pin ts-patch for api build

lucasmarshall:


apps/api/Dockerfile
```diff
@@ -23,7 +23,7 @@ COPY turbo.json turbo.json
 COPY tsconfig.json tsconfig.json
 
 RUN --mount=type=cache,mode=0777,target=/app/.yarn/cache yarn workspace @supaglue/db prisma generate
-RUN --mount=type=cache,mode=0777,target=/app/.yarn/cache yarn workspace api dlx ts-patch install -s
+RUN --mount=type=cache,mode=0777,target=/app/.yarn/cache yarn workspace api dlx -p ts-patch@2.1.0 ts-patch install -s
```
lucasmarshall, line 26: 3.0 breaks

apps/salesforce-pub-sub/Dockerfile
```diff
@@ -23,7 +23,6 @@ COPY turbo.json turbo.json
 COPY tsconfig.json tsconfig.json
 
 RUN --mount=type=cache,mode=0777,target=/app/.yarn/cache yarn workspace @supaglue/db prisma generate
-RUN --mount=type=cache,mode=0777,target=/app/.yarn/cache yarn workspace salesforce-pub-sub dlx ts-patch install -s
```
lucasmarshall, line 26: Not used in this service

apps/api/Dockerfile
albertyfwu: what's the way to make this work with 3.0?

albertyfwu (approved):

apps/api/Dockerfile
lucasmarshall: https://github.com/supaglue-labs/supaglue/pull/1003

PR #1007, opened 2023-06-13
fix: underscored ids for certain applications for v2 syncs

asdfryan:
Tested locally

Screenshot 2023-06-13 at 5 15 11 PM


lucasmarshall (commented):
Why is this needed? Why can't we have the same schema in all applications?

asdfryan:
> Why is this needed? Why can't we have the same schema in all applications?

I think it's not needed anymore. It's a bit of a long story, but it was previously thought that Warmly needed the `id` field to be reserved for something else due to some codegen tool they were using. But now it seems like they've discovered a workaround.

PR #1019, opened 2023-06-15
feat(mgmt-ui): add destination testing

tomkit:
- server-side test then save flow for postgres and s3

https://github.com/supaglue-labs/supaglue/assets/471516/63a53ccf-42cc-45d7-969a-61b39c23170d

https://github.com/supaglue-labs/supaglue/assets/471516/71cf4d82-7d49-4a8a-98e0-31e66257a593

## Test Plan

[Describe test plan here]

## Deployment instructions


packages/schemas/v2/mgmt.ts
```diff
@@ -32,6 +32,11 @@ export type CreateDestinationRequest = operations['createDestination']['requestB
 export type CreateDestinationResponse =
   operations['createDestination']['responses'][keyof operations['createDestination']['responses']]['content']['application/json'];
 
+// Internal API only so no generated OpenAPI types:
```
albertyfwu, line 35: if there are no generated types, then why not just leave it untyped? `packages/schemas/v2/mgmt.ts` would suggest that this only contains types for mgmt routes

packages/core/services/destination_service.ts
```diff
@@ -59,6 +68,66 @@ export class DestinationService {
...
+              accessKeyId: params.config.accessKeyId,
+              secretAccessKey: params.config.secretAccessKey,
+            },
+            region: params.config.region,
+            requestHandler: new NodeHttpHandler({
+              connectionTimeout: 1500,
+            }),
+          });
+
+          const command = new ListBucketsCommand({}); // Use listing as a proxy for appropriate credentials
```
albertyfwu, line 89: can this be expensive if there are a lot of buckets? Is there some query that is cheaper?

albertyfwu (approved):

packages/core/services/destination_service.ts
asdfryan, line 89: AWS has different permissions for `ListBucket` and `WriteBucket` afaik. Just because you can call `List` doesn't mean you can do writes.

tomkit: Understood: the same is true for the Postgres check. I think if we wanted to do a realistic check we would have to do writes (in S3) and writes and table creation (in Postgres) which I want to avoid for now

tomkit: I'll see if there's a way to limit it or find another command that is overall, on average, cheaper

packages/schemas/v2/mgmt.ts
tomkit: - mgmt and internal endpoint types are tightly coupled right now so just keeping that consistent until we change that pattern
- i don't do any client-side validation so relying on server-side to catch customer-facing validation

albertyfwu: only openapi middleware does validation. i don't think we have openapi middleware for internal routes. typescript types don't exist at runtime

PR #1020, opened 2023-06-15
feat: basic syncs for MS Dynamics 365

lucasmarshall:
Fixes: #1016 

## Test Plan

Tested locally


packages/core/remotes/crm/ms_dynamics_365_sales/index.ts
```diff
@@ -1,55 +1,340 @@
...
+    const url = new URL(`${this.baseUrl}${entity}`);
+    // we can't use URLSearchParams because odata doesn't expect the keys to be urlencoded
+    url.search = Object.entries(params)
+      .map(([key, value]) => {
+        if (value !== undefined && value !== null && value !== '') {
+          return `${key}=${encodeURIComponent(value)}`;
+        }
+      })
+      .filter(Boolean)
+      .join('&');
```
lucasmarshall, line 106: With some testing, it doesn't look like this is working correctly. Will fix

apps/api/routes/oauth.ts
```diff
@@ -188,10 +188,12 @@ export default function init(app: Router): void {
       // TODO: implement code_verifier/code_challenge when we implement sessions
       const additionalAuthParams: Record<string, string> = {};
 
+      const scopes = authorizeWithScope ? [...scope.split(' '), ...(additionalScopes ?? [])] : undefined;
+
       const tokenWrapper = await client.getToken({
         code,
         redirect_uri: REDIRECT_URI,
-        scope: auth.authorizeWithScope ? scope : undefined,
+        scope: scopes,
```
tomkit, line 196: Can you test this against existing providers? The reason why I had the conditional was because Salesforce doesn't allow the scope param to be present when authenticating via `/auth`

apps/api/routes/oauth.ts
```diff
@@ -167,12 +172,7 @@ export default function init(app: Router): void {
         auth.authorizeHost = loginUrl;
       }
 
-      if (loginParams) {
-        auth.tokenPath = `${auth.tokenPath}?${loginParams}`;
-        auth.authorizePath = `${auth.authorizePath}?${loginParams}`;
```
tomkit, line 172: When going through their Postman docs, they have `resource={dynamics hostname}` as a query parameter which is why I had this. https://learn.microsoft.com/en-us/power-apps/developer/data-platform/webapi/setup-postman-environment

Is there any alternate way of specifying which dynamics domain to auth for?

apps/api/routes/oauth.ts
```diff
@@ -22,7 +22,7 @@ export default function init(app: Router): void {
       req: Request<never, any, never, any, { applicationId: string; customerId: string; providerName: string }>,
       res: Response
     ) => {
-      const { applicationId, customerId, providerName, returnUrl, loginUrl, loginParams, version = 'v2' } = req.query;
+      const { applicationId, customerId, providerName, returnUrl, loginUrl, version = 'v2', scope } = req.query;
```
tomkit, line 25: Can we name it `additionalScopes` or something of that nature since it is not replacing the scopes already defined in the integration configuration

packages/core/remotes/crm/ms_dynamics_365_sales/index.ts
tomkit: Could we use `query-string` lib for this?

apps/api/routes/oauth.ts
lucasmarshall: That's for the oauth2 v1 endpoint. For futureproofing I'm using the v2.0 endpoint.

apps/api/routes/oauth.ts
lucasmarshall: It's only used for dynamics and we don't allow setting scope for dynamics (since there is no scope a developer could set)

apps/api/routes/oauth.ts
```diff
@@ -51,20 +51,28 @@ export default function init(app: Router): void {
       }
 
       const { oauthScopes } = integration.config.oauth;
+
+      if (integration.providerName === 'ms_dynamics_365_sales') {
+        // For Dynamics 365, we need to get the scope from the query params
+        // this is because the scope is in the form of: https://org8d6f84ed.crm.dynamics.com/.default
```
tomkit, line 57: Is the default scope the recommended approach? So developers don't need to specify `https://admin.services.crm.dynamics.com/user_impersonation` ?

lucasmarshall: No, that scope is not required at all

lucasmarshall: The only scope needed is `https://<org>.crm.dynamics.com/.default`

packages/core/remotes/crm/ms_dynamics_365_sales/index.ts
albertyfwu: looks like we are using `const params = new URLSearchParams();` elsewhere in the code

lucasmarshall: As said in the comment in the code `URLSearchParams` doesn't work for this.

apps/api/routes/oauth.ts
asdfryan: This should be `provider` now instead of `integration` after you rebase

asdfryan:
Do you mind rebasing? There's a lot of stuff that's changed regarding `integration` / `provider` etc

openapi/v1/crm/components/schemas/objects/email_addresses.yaml
```diff
@@ -10,6 +10,7 @@ items:
       enum:
         - primary
         - work
+      nullable: true
```
asdfryan, line 13: Hmm making this nullable will affect all other providers -- why not just make them primary?

lucasmarshall: We can make them all primary I guess. Is that what you mean?

asdfryan: I defer to you on that decision. We just need to make sure that if we make it nullable we need to be able to handle that payload for all of our crm providers. (or just keep it non-nullable)

asdfryan (approved):

openapi/v1/crm/components/schemas/objects/email_addresses.yaml
lucasmarshall: I've added an "other" type instead

PR #1031, opened 2023-06-16
fix: rename model -> record in v2 apis

asdfryan:


packages/schemas/v2/crm.ts
```diff
@@ -123,7 +123,7 @@ export type CreateCustomObjectResponse =
 
 export type GetCustomObjectPathParams = paths[`/custom-objects/{custom_object_id}`]['parameters']['path'];
 // TODO - this should be generated from the openapi spec, but it's not getting generated due to a bug in openapi-typescript
-// export type GetCustomObjectQueryParams = Required<operations['getCustomObject']>['parameters']['query'];
+// export type GetCustomObjectClassQueryParams = Required<operations['getCustomObjectClass']>['parameters']['query'];
```
albertyfwu, line 126: why change this back to objectclass?

albertyfwu (approved):

packages/schemas/v2/crm.ts
asdfryan: bad merge

PR #1032, opened 2023-06-16
feat: sync hubspot custom object without associations

albertyfwu:
Next: pull associations

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/types/sync_config.ts
+++ b/packages/types/sync_config.ts
@@ -19,8 +19,9 @@ export type SyncConfigUpdateParams = SyncConfigCreateParams;
 
 export type SyncConfigData = {
   defaultConfig: SyncStrategyConfig;
-  commonObjects: CommonObjectConfig[];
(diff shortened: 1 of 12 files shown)


packages/types/sync_config.ts
```diff
@@ -19,8 +19,9 @@ export type SyncConfigUpdateParams = SyncConfigCreateParams;
 
 export type SyncConfigData = {
   defaultConfig: SyncStrategyConfig;
-  commonObjects: CommonObjectConfig[];
-  rawObjects: RawObjectConfig[];
+  commonObjects?: CommonObjectConfig[];
```
asdfryan, line 22: Why optional?

asdfryan: I think this causes build to fail

albertyfwu: just pushed a fix

asdfryan (approved):

PR #1041, opened 2023-06-19
feat(auth): support managed oauth app

tomkit:
- add `use_managed_oauth` as a developer-customer-facing configuration
- add `applyUseManagedOauth` as an internal-facing configuration
- allow users in mgmt-ui to use managed oauth apps
- developer-customers can only set `use_managed_oauth` during create, not update
- update provider endpoints to allow `use_managed_oauth` on creation, not update
- intentionally did not update the integrations endpoints

Screenshot 2023-06-19 at 1 58 19 PM

## Test Plan

[Describe test plan here]

## Deployment instructions


packages/core/mappers/integration.ts
```diff
@@ -3,20 +3,21 @@ import {
...
-  id,
-  applicationId,
-  destinationId,
-  category,
-  providerName,
-  config,
-}: IntegrationModel): Promise<Integration> => {
+export const fromIntegrationModel = async (
+  { id, applicationId, destinationId, category, providerName, config }: IntegrationModel,
+  args: IntegrationMapperArgs = { applyUseManagedOauth: true }
```
asdfryan, line 19: rename to stripOauthConfig and reverse

packages/types/integration.ts
```diff
@@ -16,22 +16,43 @@ export type CRMIntegration = BaseIntegration & {
...
+  oauth: OAuthConfigDecrypted;
+  sync: IntegrationSyncConfig;
+  useManagedOauth: true;
+};
+
 export type IntegrationConfigDecrypted = {
   providerAppId: string;
   oauth: OAuthConfigDecrypted;
   sync: IntegrationSyncConfig;
+  useManagedOauth?: boolean;
```
asdfryan, line 48: useManagedOauth?: false

packages/types/integration.ts
```diff
@@ -16,22 +16,43 @@ export type CRMIntegration = BaseIntegration & {
...
 };
 
 export type IntegrationConfigEncrypted = {
   providerAppId: string;
   oauth: OAuthConfigEncrypted;
   sync: IntegrationSyncConfig;
+  useManagedOauth?: boolean;
```
asdfryan, line 55: useManagedOauth?: false

apps/api/routes/internal/integration/index.ts
```diff
@@ -31,7 +31,9 @@ export default function init(app: Router): void {
       req: Request<GetIntegrationsPathParams, GetIntegrationsResponse, GetIntegrationsRequest>,
       res: Response<GetIntegrationsResponse>
     ) => {
-      const integrations = await integrationService.list(req.supaglueApplication.id);
+      const integrations = await integrationService.list(req.supaglueApplication.id, {
```
asdfryan, line 34: we should also update the mgmt API

packages/core/mappers/integration.ts
tomkit: it empties the field value, but not the key, named it `hideManagedOauthConfig`

packages/types/integration.ts
tomkit: the oauth config signature doesn't change based on this, let's keep it loose so we don't need to do a type tree shake

asdfryan (approved):
This seems fine to me, but you'll have to re-implement the FE after my PR I think.

PR #1048, opened 2023-06-20
fix: posthog events for oauth and sync

lucasmarshall:
## Test Plan

Tested locally


packages/sync-workflows/services/sync_service.ts
albertyfwu, line 1: nit (optional): yea, `sync-worker` was probably not the right place to put these services.

However, if we're going to move them into `sync-workflows`, they might as well be in `@supaglue/core`, since both `sync-workflows` and `core` are shared by both `api` and `sync-worker`.

albertyfwu (commented):
Looks like there are some type/build issues but otherwise lgtm

packages/sync-workflows/services/sync_service.ts
lucasmarshall: It actually looks like `@supaglue/core` has it's own implementations of these, so I'll keep them here for now.

apps/api/package.json
```diff
@@ -14,6 +14,7 @@
     "@supaglue/sync-workflows": "workspace:*",
     "@temporalio/activity": "^1.6.0",
     "@temporalio/client": "^1.6.0",
+    "@temporalio/worker": "^1.6.0",
```
lucasmarshall, line 17: Now a peer dependency for `@supaglue/sync-workflows`

albertyfwu (approved):

apps/api/package.json
albertyfwu: actually i think this was the main reason i didn't wnat it in `sync-workflows`. there's no reason for `api` to depend on `@temporalio/worker`

lucasmarshall: There's no way we can use it in `@supaglue/sync-workflows` as we do without it either being a dependency or a peer dependency of that. Either way, we'll be installing it in `api` even though it isn't used.

I suppose we could possibly move the interceptor stuff to `sync-worker`, removing it from `@supaglue/sync-workflows` and then we could remove this dependency from `api`

albertyfwu: i'd say:
- if it's too much work leave it as-is,
- slightly harder solution is probably to migrate these services to `core`,
- even harder solution is moving interceptor to `sync-worker`, etc.

probably not worth spending too much time on it if we have other priorities

lucasmarshall: It seems that the only code in `@supaglue/sync-workflows` that used this wasn't used at all, so I've removed it and updated the dependencies

package.json
```diff
@@ -44,19 +44,12 @@
...
     "tsc-files": "^1.1.3",
     "tsx": "^3.12.3",
-    "turbo": "^1.10.1",
+    "turbo": "^1.10.3",
     "typescript": "^4.9.5"
   },
   "engines": {
     "node": "18.x"
   },
-  "dependencies": {
```
asdfryan, line 53: Why did we remove this? @lucasmarshall

lucasmarshall: The workspace should not have any dependencies of its own.

PR #1053, opened 2023-06-21
chore: remove v1 endpoints and openapi specs

asdfryan:
Keeping around `crm/v1/passthrough` since it's still being used by a customer


apps/api/routes/crm/v1/index.ts
```diff
@@ -1,23 +1,10 @@
...
-
-  contacts(v1Router);
-  accounts(v1Router);
-  leads(v1Router);
-  opportunities(v1Router);
-  users(v1Router);
-
   passthrough(v1Router);
 
   v1Router.use(openApiErrorHandlerMiddleware);
```
lucasmarshall, line 10: We don't need this middleware if we aren't using the openapi middleware as well

lucasmarshall (changes requested):

lucasmarshall (approved):

PR #1060, opened 2023-06-21
chore: oncall alert cleanup

tomkit:
- add `environment` field to application
- lower webhook from error to warn
- warn on sync failures for `environment` === `development` (i'll set this for certain customer applications)

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/lib/webhook.ts
+++ b/packages/core/lib/webhook.ts
@@ -30,7 +30,7 @@ export const maybeSendWebhookPayload = async (
     });
   } catch (e) {
     // TODO: Don't swallow this error.
-    logger.error(e, 'Failed to send webhook');
+    logger.warn(e, 'Failed to send webhook');
   }
 };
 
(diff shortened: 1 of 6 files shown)


packages/core/lib/webhook.ts
```diff
@@ -30,7 +30,7 @@ export const maybeSendWebhookPayload = async (
     });
   } catch (e) {
     // TODO: Don't swallow this error.
-    logger.error(e, 'Failed to send webhook');
+    logger.info(e, 'Failed to send webhook');
```
lucasmarshall, line 33: Use warn instead?

packages/sync-workflows/activities/log_sync_finish.ts
```diff
@@ -32,14 +32,21 @@ export function createLogSyncFinish({
...
+
+    const connection = await connectionService.getSafeById(connectionId);
+    const application = await applicationService.getById(connection.applicationId);
+
     if (status === 'FAILURE') {
       const error = new Error(errorMessage);
       error.stack = errorStack;
-      logger.error(error, `Sync failed for syncId ${syncId} and connectionId ${connectionId}`);
+
+      if (application.environment === 'development') {
```
lucasmarshall, line 43: Do we want to swallow these errors for self-hosted as well? Or just when `IS_CLOUD` is set?

lucasmarshall (changes requested):

lucasmarshall (approved):

PR #1062, opened 2023-06-22
chore: add field mapping for syncs

albertyfwu:
- Developer can define schema
- Customer can choose how to map to schema
- Developer can enable/disable sync (distinct from pause/unpause... that isn't implemented yet)

Example config for `SyncConfig`:

```
{
    "application_id": "f72c1b96-38fc-42ca-a82a-68e81f3cbebc",
    "provider_id": "c92b0374-7c4f-42fe-9506-65c7e297dcbc",
    "destination_id": "59ab3dc1-a51a-4bb8-bf36-99c0c4757713",
    "config": {
        "common_objects": [],
        "default_config": {
            "period_ms": 3600000,
            "strategy": "full then incremental"
        },
        "standard_objects": [
            {
                "object": "company",
                "schema": {
                    "type": "defined",
                    "fields": [
                        {
                            "name": "new_name",
                            "mapped_name": "name"
                        },
                        {
                            "name": "new_city",
                            "mapped_name": "city"
                        },
                        {
                            "name": "best_time"
                        }
                    ],
                    "allow_additional_field_mappings": true
                }
            }
        ],
        "custom_objects": [
            {
                "object": "trucks"
            }
        ]
    }
}
```

Example config for `Sync`'s `schema_mappings_config`:
```
{
    "schema_mappings_config": {
        "standard_objects": [
            {
                "object": "company",
                "field_mappings": [
                    {
                        "schema_field": "best_time",
                        "mapped_field": "hs_lastmodifieddate"
                    }
                ]
            }
        ]
    }
}
```

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/api/routes/internal/sync_config/index.ts
+++ b/apps/api/routes/internal/sync_config/index.ts
@@ -35,11 +35,12 @@ export default function init(app: Router): void {
       const syncConfigs = await syncConfigService.list(req.supaglueApplication.id);
       return res.status(200).send(
         syncConfigs.map((syncConfig) => {
-          const { rawObjects, rawCustomObjects, ...rest } = syncConfig.config;
+          const { defaultConfig, rawObjects, rawCustomObjects, commonObjects } = syncConfig.config;
           return snakecaseKeys({
             ...syncConfig,
             config: {
-              ...rest,
+              defaultConfig,
+              commonObjects,
               standardObjects: rawObjects, // TODO: remove when we propogate `standardObjects` all the way through code
               customObjects: rawCustomObjects,
             },
@@ -55,18 +56,19 @@ export default function init(app: Router): void {
       req: Request<CreateSyncConfigPathParams, CreateSyncConfigResponse, CreateSyncConfigRequest>,
       res: Response<CreateSyncConfigResponse>
     ) => {
+      const { default_config, common_objects, standard_objects, custom_objects } = req.body.config;
       const syncConfig = await syncConfigService.create({
         applicationId: req.supaglueApplication.id,
         ...camelcaseKeys({
           ...req.body,
           config: {
-            ...req.body.config,
-            common_objects: req.body.config.common_objects?.map((commonObject) => ({
+            default_config,
+            common_objects: common_objects?.map((commonObject) => ({
               ...commonObject,
               object: commonObject.object as CommonModelType,
             })),
-            raw_objects: req.body.config.standard_objects,
-            raw_custom_objects: req.body.config.custom_objects,
+            raw_objects: standard_objects,
+            raw_custom_objects: custom_objects,
           },
         }),
       });
@@ -84,12 +86,13 @@ export default function init(app: Router): void {
         req.params.sync_config_id,
         req.supaglueApplication.id
       );
-      const { rawObjects, rawCustomObjects, ...rest } = syncConfig.config;
+      const { defaultConfig, commonObjects, rawObjects, rawCustomObjects } = syncConfig.config;
       return res.status(200).send(
         snakecaseKeys({
           ...syncConfig,
           config: {
-            ...rest,
+            defaultConfig,
+            commonObjects,
             standardObjects: rawObjects, // TODO: remove when we propogate `standardObjects` all the way through code
             customObjects: rawCustomObjects,
           },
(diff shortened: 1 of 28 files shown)


packages/types/sync_config.ts
```diff
@@ -23,7 +23,7 @@ export type SyncStrategyConfig = {
   strategy: SyncType;
   // When this is undefined, we treat it as "true" to be backwards compatible.
   // TODO: When we have time, migrate this to be a required field.
-  startSyncOnConnectionCreation?: boolean;
+  enableSyncOnConnectionCreation?: boolean;
```
albertyfwu, line 26: renamed this because we will want to distinguish enabled/disabled and paused/unpaused concepts cc @george-xing @asdfryan 

A developer may want to have syncs enabled or disabled for various customers, and customers (or developers) may want to pause/resume them.

Implementation-wise, that would be the difference between not creating a sync in DB, or creating a sync in DB and having it paused in temporal

packages/types/sync_config.ts
```diff
@@ -38,7 +38,7 @@ export type RawObjectConfig = {
 };
 
 export type RawObjectInheritedSchema = {
-  type: 'inherited';
+  type: 'inherit_all_fields';
```
albertyfwu, line 41: clearer?

albertyfwu: not relevant anymore

packages/types/sync_config.ts
albertyfwu: @george-xing @tomkit  any thoughts on the terminology: `enable/disable` or `create/delete` a sync? I was thinking of implementing

`POST /.../customers/{customer_id}/connections/{connection_id}/sync` to enable (or create) a sync, and
`DELETE /../customers/{customer_id}/connections/{connection_id/sync` to disable (or delete) a sync, and

for pausing/resuming, it would updating some property on the result of
`GET /../customers/{customer_id}/connections/{connection_id}/sync`, like
`PATCH /../customers/{customer_id}/connections/{connection_id}/sync` or
`POST /../customers/{customer_id}/connections/{connection_id}/sync/_pause`

apps/api/routes/internal/sync_config/index.ts
```diff
@@ -35,11 +35,12 @@ export default function init(app: Router): void {
       const syncConfigs = await syncConfigService.list(req.supaglueApplication.id);
       return res.status(200).send(
         syncConfigs.map((syncConfig) => {
-          const { rawObjects, rawCustomObjects, ...rest } = syncConfig.config;
```
albertyfwu, line 38: this was causing our differently-named objects from the DB to be leaked to the API caller

packages/types/sync_config.ts
albertyfwu: i decided on `enable/disable`

tomkit: sgtm, names sound good and also the rest for crud, POST for rpc-like operations

tomkit (approved):

packages/types/sync_config.ts
asdfryan: If you disable a sync and then re-enable it, will the `syncId` be guaranteed to be the same? If so I'm fine with this.

Because `disabled` and `enabled` are really properties of the sync, I would expect the `syncId` to be stable.

packages/sync-workflows/activities/sync_raw_records_to_destination.ts
```diff
@@ -62,14 +53,39 @@ export function createSyncRawRecordsToDestination(
...
+              return await (client as CrmRemoteClient).listRawStandardObjectRecords(
+                object,
+                fieldMappingConfig,
+                modifiedAfter,
+                heartbeat
+              );
+            })();
+        return await writer.writeRawRecords(connection, object, toHeartbeatingReadable(stream), heartbeat);
+      } else {
+        throw ApplicationFailure.nonRetryable(`Unsupported category: ${client.category()}`);
```
asdfryan, line 86: Can you be more specific with the message? Like `syncing standard objects is only supported for the crm category` or something.

packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -385,10 +390,39 @@ class HubSpotClient extends AbstractCrmRemoteClient {
       modifiedAfter
     );
 
+    // Create map from mappedField -> schemaField
```
asdfryan, line 393: Does this work if one of the mapped fields is the same as one of the (previously) unmapped fields?

packages/sync-workflows/activities/sync_raw_records_to_destination.ts
albertyfwu: ok

packages/core/remotes/crm/hubspot/index.ts
albertyfwu: yea there is an issue here if
colA -> hubspot_1
colB -> hubspot_2
i'll follow up with a fix

PR #1064, opened 2023-06-22
fix: propagate BE errors to FE and notify

asdfryan:
Fixes: #### 

After this change we no longer ignore BE errors and propagate them to the FE.

## Test Plan

Testing locally

image

## Deployment instructions


apps/mgmt-ui/src/pages/api/internal/webhook/create.ts
```diff
@@ -10,11 +10,10 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse<
     body: JSON.stringify(req.body),
   });
 
+  const r = await result.json();
```
albertyfwu, line 13: will this always succeed even when `result.ok === false`?

tomkit (approved):

apps/mgmt-ui/src/pages/api/internal/webhook/create.ts
asdfryan: Yes.

PR #1075, opened 2023-06-22
chore: change messaging to say active connections

asdfryan:


packages/core/services/sync_config_service.ts
```diff
@@ -141,7 +141,7 @@ export class SyncConfigService {
       },
     });
     if (syncs.length) {
-      throw new BadRequestError('Cannot delete sync config with active syncs');
+      throw new BadRequestError('Cannot delete sync config with active connections');
```
albertyfwu, line 144: this is also kind of unclear; why is a connection "active"? but maybe this is OK for now until we actually introduce `sync` as first-class concept on mgmt-ui

albertyfwu (approved):

PR #1078, opened 2023-06-22
fix: misc ui fixes

asdfryan:
Screenshot 2023-06-22 at 2 23 06 PM

### DIFF
--- a/apps/mgmt-ui/src/components/configuration/syncConfig/SyncConfigListPanel.tsx
+++ b/apps/mgmt-ui/src/components/configuration/syncConfig/SyncConfigListPanel.tsx
@@ -123,7 +123,6 @@ export default function SyncConfigListPanel() {
     frequency: `every ${millisToHumanReadable(syncConfig.config.defaultConfig.periodMs)}`,
     objects: getObjectsString(syncConfig),
   }));
-  console.log(`rows: `, rows);
 
   return (
     <div className="flex flex-col gap-4">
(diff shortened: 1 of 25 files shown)


apps/mgmt-ui/src/components/configuration/syncConfig/SyncConfigListPanel.tsx
```diff
@@ -203,3 +205,20 @@ function millisToHumanReadable(millis: number): string {
 
   return humanReadable;
 }
+
+function getObjectsString(syncConfig: SyncConfig): string {
```
tomkit, line 209: we had an in-office discussion about separating the sync logs into {standard objects, other objects}. can we do that here too to be consistent from a ui perspective?

asdfryan: These aren't sync logs, this is just a short form list of objects that are being synced for the list view for sync configs.

tomkit (approved):
ideally we can have the columns be {standard objects, objects} so it matches the sync log information layout

apps/mgmt-ui/src/components/configuration/syncConfig/SyncConfigListPanel.tsx
albertyfwu: i'm doing the "separating the sync logs into {standard objects, other objects}"

albertyfwu: https://github.com/supaglue-labs/supaglue/pull/1080

albertyfwu (approved):

PR #1093, opened 2023-06-22
fix: support custom object in dynamics

lucasmarshall:


asdfryan (approved):

packages/core/remotes/impl/ms_dynamics_365_sales/index.ts
```diff
@@ -71,7 +76,7 @@ class MsDynamics365Sales extends AbstractCrmRemoteClient {
     'OData-MaxVersion': '4.0',
     'OData-Version': '4.0',
     'Content-Type': 'application/json',
-    Prefer: `odata.maxpagesize=${MAX_PAGE_SIZE}`,
+    Prefer: `odata.maxpagesize=${MAX_PAGE_SIZE},return=representation`,
```
asdfryan, line 79: I think this breaks existing common objects

PR #1110, opened 2023-06-26
fix: sort hubspot search by id for more stable sort order

asdfryan:
Tested locally


packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -1284,7 +1284,7 @@ class HubSpotClient extends AbstractCrmRemoteClient {
         ],
         sorts: [
           {
-            propertyName: 'lastmodifieddate',
+            propertyName: 'id',
```
albertyfwu, line 1287: can you write comments on all of these talking about the reasoning? otherwise we surely will not remember

albertyfwu (approved):

PR #1113, opened 2023-06-26
feat: actions api for MS Dynamics 365

lucasmarshall:
## Test Plan

Tested locally against the trial dynamics instance

### DIFF
--- a/packages/core/remotes/crm/ms_dynamics_365_sales/index.ts
+++ b/packages/core/remotes/crm/ms_dynamics_365_sales/index.ts
@@ -34,6 +34,14 @@ import {
   fromDynamicsLeadToRemoteLead,
   fromDynamicsOpportunityToRemoteOpportunity,
   fromDynamicsUserToRemoteUser,
+  toDynamicsAccountCreateParams,
+  toDynamicsAccountUpdateParams,
+  toDynamicsContactCreateParams,
+  toDynamicsContactUpdateParams,
+  toDynamicsLeadCreateParams,
+  toDynamicsLeadUpdateParams,
+  toDynamicsOpportunityCreateParams,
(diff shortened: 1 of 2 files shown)


packages/core/remotes/crm/ms_dynamics_365_sales/index.ts
```diff
@@ -62,7 +70,7 @@ class MsDynamics365Sales extends AbstractCrmRemoteClient {
     super(`${credentials.instanceUrl}api/data/v9.2/`);
     this.#credentials = credentials;
     this.#headers.Authorization = `Bearer ${this.#credentials.accessToken}`;
-    this.#odata = o(this.baseUrl, { headers: this.#headers });
+    this.#odata = o(this.baseUrl, { headers: this.#headers, referrer: undefined });
```
lucasmarshall, line 73: This needs to be set to undefined for the odata client to work with the built in node fetch method.

packages/core/remotes/crm/ms_dynamics_365_sales/index.ts
```diff
@@ -190,46 +199,94 @@ class MsDynamics365Sales extends AbstractCrmRemoteClient {
     commonModelType: T,
     params: CRMCommonModelTypeMap<T>['createParams']
   ): Promise<string> {
-    throw new Error('Not implemented');
+    await this.maybeRefreshAccessToken();
+    switch (commonModelType) {
+      case 'account': {
+        const response = await this.#odata.post('accounts', toDynamicsAccountCreateParams(params)).query();
+        const id = response.headers.get('location')?.split('(')[1].split(')')[0];
```
albertyfwu, line 206: don't we need to use `?` on every accessor after the first one?

albertyfwu (approved):

PR #1118, opened 2023-06-26
chore: upgrade prisma

lucasmarshall:


packages/db/package.json
```diff
@@ -6,12 +6,12 @@
   "index": "dist/index.js",
   "types": "index.ts",
   "dependencies": {
-    "@prisma/client": "^4.12.0",
+    "@prisma/client": "^4.16.1",
```
lucasmarshall, line 9: Looks like I forgot to upgrade this last time we upgraded

PR #1121, opened 2023-06-26
chore: fix hubspot managed oauth scopes

lucasmarshall:


albertyfwu (approved):
fwiw i don't have `oauth` scope in my local provider and i'm able to successfully sync objects

PR #1127, opened 2023-06-27
chore: schema endpoints, openapi spec, types, mappers and service

asdfryan:
This just implements the CRUD endpoints for schemas without connecting to them to `SyncConfig`s or `Connection`s.

TODOs:
- Move schema mappings from `sync` object to `connection` object
- Migrate `SyncConfig` to reference `Schema` objects instead of defining them inline


openapi/v2/mgmt/paths/schemas@{schema_id}.yaml
```diff
@@ -0,0 +1,90 @@
...
+                  allow_additional_field_mappings: false
+delete:
+  operationId: deleteSchema
+  summary: Delete schema
+  tags:
+    - Schemas
+  security:
+    - ApiKeyAuth: []
+  responses:
+    200:
```
tomkit, line 66: are most of our deletes are empty body / http 204?

tomkit (approved):

apps/api/routes/internal/schema/index.ts
```diff
@@ -0,0 +1,89 @@
+import { getDependencyContainer } from '@/dependency_container';
```
albertyfwu, line 1: nit: if our internal and mgmt endpoints always look identical, we should probably just write the code once and share it

albertyfwu (approved):

packages/types/schema.ts
```diff
@@ -0,0 +1,19 @@
...
+  allowAdditionalFieldMappings: boolean;
+};
+
+export type SchemaField = {
+  name: string; // my_first_column
+  mappedName?: string; // salesforce_first_column
+};
+
+export type SchemaCreateParams = Omit<Schema, 'id'>;
+export type SchemaUpdateParams = SchemaCreateParams;
```
albertyfwu, line 19: `applicationId` should not be changeable

PR #1128, opened 2023-06-27
fix: make all postgres table names lowercase

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


asdfryan (commented):
We should validate that no two objects write to the same place (i.e. `Contact` vs `contact`)

lucasmarshall (approved):

albertyfwu:
> We should validate that no two objects write to the same place (i.e. `Contact` vs `contact`)

there is a lot of validation we should do. there is a separate task for that on our task board.

PR #1135, opened 2023-06-27
docs: some cleanup

tomkit:
- Add reference to read pattern in action api references
- Convert common schema to common object
- Associate integration patterns with reads and writes

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/docs/docs/integration-patterns/managed-syncs.md
+++ b/docs/docs/integration-patterns/managed-syncs.md
@@ -2,7 +2,7 @@
 description: ''
 ---
 
-# Managed syncs
+# Managed syncs (reads)
 
 Managed Syncs lets you sync data from your customers’ CRM directly into your own application database or data warehouse.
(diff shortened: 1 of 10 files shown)


docs/docs/integration-patterns/managed-syncs.md
```diff
@@ -2,7 +2,7 @@
 description: ''
 ---
 
-# Managed syncs
+# Managed syncs (reads)
```
lucasmarshall, line 5: This is a bit confusing. It's both a read (from the provider) and a write (to the developer's db)

lucasmarshall (changes requested):

docs/docs/integration-patterns/managed-syncs.md
tomkit: From a technology standpoint, but from a use-case and product standpoint the developer is trying to read provider data

lucasmarshall (approved):

PR #1139, opened 2023-06-28
chore: implement object syncs

albertyfwu:
This implements object syncs (one Temporal schedule per object).

This is implemented parallel to the existing system, and is not active. The next PR will include code to migrate to the new system, etc.

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/mappers/object_sync_run.ts
+++ b/packages/core/mappers/object_sync_run.ts
@@ -1,0 +1,24 @@
+import { SyncHistoryStatus } from '@supaglue/types';
+import { ObjectSyncRun } from '@supaglue/types/object_sync_run';
+import { parseCustomerIdPk } from '../lib/customer_id';
+import { ObjectSyncRunModelExpanded } from '../types/object_sync_run';
+
+export const fromObjectSyncRunModelAndSync = (args: ObjectSyncRunModelExpanded): ObjectSyncRun => {
+  const { id, status, errorMessage, startTimestamp, endTimestamp, objectSync, numRecordsSynced } = args;
+  const { connection } = objectSync;
+  const { applicationId, externalCustomerId } = parseCustomerIdPk(connection.customerId);
+  return {
+    id,
+    objectSyncId: objectSync.id,
+    status: status as SyncHistoryStatus,
+    errorMessage,
+    startTimestamp,
+    endTimestamp,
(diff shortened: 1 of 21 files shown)


packages/core/mappers/object_sync_run.ts
```diff
@@ -0,0 +1,24 @@
+import { SyncHistoryStatus } from '@supaglue/types';
```
lucasmarshall, line 1:
```suggestion
import type { SyncHistoryStatus } from '@supaglue/types';
```

packages/core/mappers/object_sync_run.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ObjectSyncRun } from '@supaglue/types/object_sync_run';
```
lucasmarshall, line 2:
```suggestion
import type { ObjectSyncRun } from '@supaglue/types/object_sync_run';
```

packages/core/mappers/object_sync_run.ts
```diff
@@ -0,0 +1,24 @@
...
+import { parseCustomerIdPk } from '../lib/customer_id';
+import { ObjectSyncRunModelExpanded } from '../types/object_sync_run';
```
lucasmarshall, line 4:
```suggestion
import type { ObjectSyncRunModelExpanded } from '../types/object_sync_run';
```

packages/core/services/object_sync_run_service.ts
```diff
@@ -0,0 +1,162 @@
+import { ObjectSyncRunModelExpanded } from '@supaglue/core/types/object_sync_run';
+import { PrismaClient } from '@supaglue/db';
```
lucasmarshall, line 2:
```suggestion
import type { ObjectSyncRunModelExpanded } from '@supaglue/core/types/object_sync_run';
import type { PrismaClient } from '@supaglue/db';
```

... etc.

lucasmarshall (approved):
Just some nits. When we are just using types from a module we should use `import type`. This will reduce runtime dependencies and may prevent unintended import-time side-effects.

albertyfwu:
> Just some nits. When we are just using types from a module we should use `import type`. This will reduce runtime dependencies and may prevent unintended import-time side-effects.

done

PR #1140, opened 2023-06-28
chore: move object schema mappings to provider

asdfryan:


apps/api/routes/internal/provider/index.ts
```diff
@@ -17,6 +17,7 @@ import {
   UpdateProviderRequest,
   UpdateProviderResponse,
 } from '@supaglue/schemas/v2/mgmt';
+import { ProviderCreateParams } from '@supaglue/types';
```
lucasmarshall, line 20:
```suggestion
import type { ProviderCreateParams } from '@supaglue/types';
```

apps/api/routes/mgmt/v2/provider/index.ts
```diff
@@ -17,6 +17,7 @@ import {
   UpdateProviderRequest,
   UpdateProviderResponse,
 } from '@supaglue/schemas/v2/mgmt';
+import { ProviderCreateParams } from '@supaglue/types';
```
lucasmarshall, line 20:
```suggestion
import type { ProviderCreateParams } from '@supaglue/types';
```

lucasmarshall (approved):
Looks ok, just a couple of nits. If all we are doing is importing a type from a module, we should use `import type`. Doing this consistently will reduce runtime dependencies and can prevent unintended import-time side effects.

PR #1146, opened 2023-06-28
fix: apply schema to actions api

asdfryan:
- Apply the object schema (if given) to the actions API

## Test Plan

Tested locally

## Deployment instructions

### DIFF
--- a/packages/core/remotes/crm/base.ts
+++ b/packages/core/remotes/crm/base.ts
@@ -32,7 +32,8 @@ export interface CrmRemoteClient extends RemoteClient {
 
   getCommonObjectRecord<T extends CRMCommonModelType>(
     commonModelType: T,
-    id: string
+    id: string,
+    fieldMappingConfig: FieldMappingConfig
   ): Promise<CRMCommonModelTypeMap<T>['object']>;
   createCommonObjectRecord<T extends CRMCommonModelType>(
     commonModelType: T,
@@ -95,7 +96,8 @@ export abstract class AbstractCrmRemoteClient extends AbstractRemoteClient imple
   ): Promise<Readable>;
   abstract getCommonObjectRecord<T extends CRMCommonModelType>(
     commonModelType: T,
-    id: string
+    id: string,
+    fieldMappingConfig: FieldMappingConfig
   ): Promise<CRMCommonModelTypeMap<T>['object']>;
   abstract createCommonObjectRecord<T extends CRMCommonModelType>(
     commonModelType: T,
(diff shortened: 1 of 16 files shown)


asdfryan:
I'm not sure why the docs build is broken -- it looks like it's using an old version of our prisma schemas

lucasmarshall (approved):

packages/core/remotes/crm/base.ts
```diff
@@ -32,7 +32,8 @@ export interface CrmRemoteClient extends RemoteClient {
 
   getCommonObjectRecord<T extends CRMCommonModelType>(
     commonModelType: T,
-    id: string
+    id: string,
+    fieldMappingConfig: FieldMappingConfig
```
albertyfwu, line 36: long term it would be nice to not have to pass this into every remote client. i think we should be able to apply the mappings afterwards, but we can leave that as tech debt. i think it would be much less code to implement per remote client

packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -969,11 +970,14 @@ class HubSpotClient extends AbstractCrmRemoteClient {
...
     await this.maybeRefreshAccessToken();
     const company = await this.#client.crm.companies.basicApi.getById(id, properties);
-    return fromHubSpotCompanyToAccount(company);
+    return {
+      ...fromHubSpotCompanyToAccount(company),
+      rawData: toMappedProperties(company.properties, fieldMappingConfig),
+    };
   }
 
   public async createAccount(params: AccountCreateParams): Promise<string> {
```
albertyfwu, line 983: can't customer pass in custom fields when creating/updating? do those need to be mapped?

asdfryan: Yes. We are only doing the response mapping, and not the request mapping. I'll file a ticket for this.

packages/core/remotes/crm/base.ts
asdfryan: That's a fair point. I'll make a ticket for this.

packages/core/remotes/crm/hubspot/index.ts
asdfryan: https://linear.app/supaglue/issue/SUP1-410/apply-schema-mapping-to-requests-for-actions-api

packages/core/remotes/crm/base.ts
asdfryan: https://linear.app/supaglue/issue/SUP1-411/remove-fieldmappingconfig-from-remote-client

PR #1149, opened 2023-06-29
feat: provider POST /object endpoint

asdfryan:
Convenience endpoint to:
- add an object to a provider
- (optionally) attach a schema to that provider
- (optionally) add that object to the syncconfig

## Test Plan

Tested locally (works)

## Deployment instructions


openapi/v2/mgmt/components/schemas/add_object.yaml
```diff
@@ -0,0 +1,26 @@
...
+      name:
+        type: string
+        example: object_name
+      type:
+        type: string
+        enum:
+          - common
+          - standard
+          - custom
+      enable_sync:
```
albertyfwu, line 13: why is this in this endpoint and not on the syncconfig? is this a shortcut/convenience feature?

apps/api/routes/mgmt/v2/provider/index.ts
```diff
@@ -52,6 +55,19 @@ export default function init(app: Router): void {
     }
   );
 
+  providerRouter.post(
+    '/:provider_id/add_object',
```
albertyfwu, line 59: i thought we have a convention for prefixing non-REST-ful interfaces with underscore. wdyt?

also, is an alternative path `'/:provider_id/objects'`?

packages/core/mappers/provider.ts
```diff
@@ -48,7 +48,7 @@ export const fromProviderModel = async ({
     config: await fromProviderConfigModel(config, {
       managedOauthConfig: managedOAuthConfigs[name],
     }),
-    objects,
+    objects: objects ? objects : undefined,
```
albertyfwu: nit: `objects ?? undefined`

packages/core/services/provider_service.ts
```diff
@@ -65,6 +75,56 @@ export class ProviderService {
     return fromProviderModel(createdProvider);
   }
 
+  public async addObject(
+    providerId: string,
+    applicationId: string,
+    params: AddObjectToProviderParams
+  ): Promise<Provider> {
+    if (params.schemaId && params.schema) {
```
albertyfwu, line 83: how is this possible? i thought the openapi spec has `oneOf` for this

openapi/v2/mgmt/components/schemas/add_object.yaml
albertyfwu: can we just leave this for syncconfig for simplicity?

asdfryan: please see https://www.notion.so/supaglue/Developer-Schema-Configuration-37ab9511e20245cdb4ae85d0dca37176?pvs=4#f78d56dc92c342ff9d714fdfa179a994 for why it's on the provider

packages/core/services/provider_service.ts
asdfryan: I changed it back because it doesn't actually work if both oneOfs are optional.

apps/api/routes/mgmt/v2/provider/index.ts
lucasmarshall: Yeah, agree with `/:provider_id/objects` as the path. That is RESTful.

lucasmarshall (approved):

PR #1152, opened 2023-06-29
fix: small fix for processSyncChanges

albertyfwu:
Currently we are early-exiting on some errs even though we call multiple Temporal APIs sequentially. We want to get through all the steps. We don't want to stop on the first failure.

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/openapi/v2/mgmt/components/schemas/sync_run.yaml
+++ b/openapi/v2/mgmt/components/schemas/sync_run.yaml
@@ -53,5 +53,6 @@ required:
   - provider_name
   - category
   - connection_id
+  - num_records_synced
   - object_type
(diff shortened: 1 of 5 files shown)


openapi/v2/mgmt/components/schemas/sync_run.yaml
```diff
@@ -53,5 +53,6 @@ required:
   - provider_name
   - category
   - connection_id
+  - num_records_synced
```
albertyfwu, line 56: `num_records_synced` is nullable, so it should be required

tomkit (approved):

PR #1157, opened 2023-06-29
feat: add endpoint for fetching all properties for a given object

asdfryan:
Addresses SUP1-393.

## Test Plan

Tested locally
Screenshot 2023-06-29 at 3 56 09 PM
Screenshot 2023-06-29 at 3 56 21 PM

## Deployment instructions


apps/api/routes/mgmt/v2/customer/connection/index.ts
```diff
@@ -35,6 +42,29 @@ export default function init(app: Router): void {
     }
   );
 
+  connectionRouter.post(
+    '/:connection_id/_list_properties',
```
albertyfwu, line 46: should we make this RESTful? GET and treat properties as first-class? or not because we don't store properties directly in our system?

packages/core/remotes/crm/hubspot/index.ts
```diff
@@ -360,7 +361,7 @@ class HubSpotClient extends AbstractCrmRemoteClient {
   }
 
   private async getStandardPropertiesToFetch(objectType: string, fieldMappingConfig?: FieldMappingConfig) {
-    const availableProperties = await this.getObjectTypeProperties(objectType);
+    const availableProperties = await this.listPropertiesForRawObjectName(objectType);
```
albertyfwu, line 364: can you remove `raw`? we moved away from `rawObjects` nomenclature

packages/types/index.ts
```diff
@@ -29,3 +29,20 @@ export type CategoryOfProviderName<T extends ProviderName> = T extends CRMProvid
 export type CommonModelForCategory<T extends ProviderCategory> = T extends 'crm'
   ? CRMCommonModelType
   : EngagementCommonModelType;
+
+export type ObjectDef<T extends ProviderCategory> = CommonObject<T> | StandardObject | CustomObject;
```
albertyfwu, line 33: do we need to do this complicated union? why not just 
```
{
  type: 'common' | 'standard' | 'custom';
  name: string;
}
```

asdfryan: Sure, we can just do the name check as application level validation.

packages/core/remotes/crm/hubspot/index.ts
asdfryan: - this is a private method so I think it doesn't really matter what it's name as long as it's clear to the developer what it means
- We need a way to distinguish between `objectName` (i.e. `account`) vs the raw object name that it is known as within the context of the provider (i.e. `company` for hubspot)

apps/api/routes/mgmt/v2/customer/connection/index.ts
asdfryan: It's already not REST-ful because this list request requires a request body

lucasmarshall: It definitely seems this should be RESTful with what we are currently putting in the body as query params

asdfryan: Done.

packages/core/remotes/crm/hubspot/index.ts
albertyfwu: can we make it `listPropertiesForCommonObject` vs `listPropertiesForObject` instead? that seems to be a little more consistent

albertyfwu (approved):

packages/core/remotes/crm/hubspot/index.ts
asdfryan: That's not quite accurate either.

The `listProperties` interface takes in an `ObjectDef` type as input, while `listPropertiesForRawObjectName` takes in the raw object name (as string) as input.

PR #1159, opened 2023-06-30
fix: emit correct errors for pipedrive

lucasmarshall:


packages/core/remotes/crm/pipedrive/index.ts
```diff
@@ -521,6 +528,29 @@ class PipedriveClient extends AbstractCrmRemoteClient {
     await this.maybeRefreshAccessToken();
     return await super.sendPassthroughRequest(request);
   }
+
+  public handleErr(err: unknown): unknown {
```
albertyfwu, line 532: can you add `override` to make it clear that this is an override?

albertyfwu (approved):

PR #1160, opened 2023-06-30
fix: null char for other call site

tomkit:
Fixes: #### 

[Describe your change here]

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/destination_writers/postgres.ts
+++ b/packages/core/destination_writers/postgres.ts
@@ -151,8 +151,9 @@ DO UPDATE SET (${columnsToUpdateStr}) = (${excludedColumnsToUpdateStr})`,
         columns,
         cast: {
           boolean: (value: boolean) => value.toString(),
-          object: (value: object) => JSON.stringify(value),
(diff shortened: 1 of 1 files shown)


asdfryan (approved):

packages/core/destination_writers/postgres.ts
```diff
@@ -151,8 +151,9 @@ DO UPDATE SET (${columnsToUpdateStr}) = (${excludedColumnsToUpdateStr})`,
         columns,
         cast: {
           boolean: (value: boolean) => value.toString(),
-          object: (value: object) => JSON.stringify(value),
+          object: (value: object) => JSON.stringify(value).replace(/\\u0000/g, ''),
```
albertyfwu, line 154: we still need to update the third callsite for cache invalidation, further up in this file

tomkit: i think that's in a different path where the developer is the source of the input and can debug + self-resolve until we add character validation for our APIs

PR #1161, opened 2023-06-30
feat: provide standard object options in UI for hubspot and salesforce

asdfryan:
Fixes SUP1-425

## Test Plan

Tested locally

https://www.loom.com/share/9c8e0514001f46759a7ca0f1a50a0633

## Deployment instructions


apps/mgmt-ui/src/components/configuration/syncConfig/SyncConfigDetailsPanel.tsx
```diff
@@ -353,3 +356,16 @@ function SyncConfigDetailsPanelImpl({ syncConfigId }: SyncConfigDetailsPanelImpl
     </div>
   );
 }
+
+const getStandardObjectOptions = (provider: Provider | undefined): string[] => {
+  switch (provider?.name) {
+    case 'hubspot': {
+      return [...HUBSPOT_STANDARD_OBJECT_TYPES];
```
albertyfwu, line 363: why do we need to make a copy?

albertyfwu (approved):

apps/mgmt-ui/src/components/configuration/syncConfig/SyncConfigDetailsPanel.tsx
asdfryan: Because the type is not `string[]`. I guess I could cast instead for performance reasons

PR #1162, opened 2023-06-30
chore: update temporal init script (namespace is required)

albertyfwu:
This just updates the `init-temporal.ts` script to pass in an additional `namespace` property that would be required for the new `temporal` binary.

Then, after a new Docker image is built for `supaglue/api`, that image (which would be used for quickstarts) would have the built `init_temporal.js` file, and we can make [another PR](https://github.com/supaglue-labs/supaglue/pull/1163) to move from temporalite to temporal.

## Test Plan

[Describe test plan here]

## Deployment instructions


apps/api/package.json
```diff
@@ -12,9 +12,9 @@
     "@supaglue/db": "workspace:*",
     "@supaglue/schemas": "workspace:*",
     "@supaglue/sync-workflows": "workspace:*",
-    "@temporalio/activity": "^1.6.0",
-    "@temporalio/client": "^1.6.0",
-    "@temporalio/workflow": "^1.6.0",
+    "@temporalio/activity": "^1.8.0",
```
albertyfwu, line 15: 1.6.0 did not have `namespace` in the gRPC schema, but the server expects it

asdfryan (approved):

PR #1164, opened 2023-06-30
fix: separate test and save buttons

asdfryan:
Fixes SUP1-409

## Test Plan

Tested locally

https://www.loom.com/share/a9a44968bfe54639b9a55d01959b45b3

## Deployment instructions

### DIFF
--- a/apps/mgmt-ui/src/components/configuration/destination/S3DestinationDetailsPanel.tsx
+++ b/apps/mgmt-ui/src/components/configuration/destination/S3DestinationDetailsPanel.tsx
@@ -268,7 +268,10 @@ export default function S3DestinationDetailsPanel({ isLoading }: S3DestinationDe
           >
             Back
           </Button>
-          {isTestSuccessful ? <SaveButton /> : <TestButton />}
+          <Stack direction="row" className="gap-2">
+            <TestButton />
+            <SaveButton />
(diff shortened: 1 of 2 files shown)


apps/mgmt-ui/src/components/configuration/destination/S3DestinationDetailsPanel.tsx
```diff
@@ -268,7 +268,10 @@ export default function S3DestinationDetailsPanel({ isLoading }: S3DestinationDe
           >
             Back
           </Button>
-          {isTestSuccessful ? <SaveButton /> : <TestButton />}
+          <Stack direction="row" className="gap-2">
+            <TestButton />
+            <SaveButton />
```
albertyfwu, line 273: shouldn't we also disable save button before test goes through successfully?

asdfryan: Yes. It is already disabled.

asdfryan: ```        disabled={!isTestSuccessful || name === '' || !isDirty} ```

albertyfwu (approved):

PR #1166, opened 2023-06-30
docs: move to using docusaurus-openapi-docs for openapi docs

lucasmarshall:


george-xing:
Thanks, could we make some changes?
1. expand the sidebar (top level) by default, and collapse the endpoints by default
2. change "API reference" page label in `/api` page to "Intro"
3. on the intro page, there's 2 version badges
4. can we put the management API first?

scripts/prepare_release.sh
```diff
@@ -53,7 +53,7 @@ done
 jq '.[0:2]' docs/versions.json > docs/versions.json.tmp && mv docs/versions.json.tmp docs/versions.json
 
 echo "Building docs..."
-yarn workspace docs build > /dev/null
+yarn build --filter "...docs" > /dev/null
```
albertyfwu, line 56: what's the difference?

albertyfwu (approved):
i'm just going to trust this does the right thing

scripts/prepare_release.sh
lucasmarshall: This uses turbo, so does caching, etc.

george-xing (approved):
Really minor - in Management API intro page, the Introduction header is bigger than the Introducer header in the other API intro pages.

lucasmarshall:
> Really minor - in Management API intro page, the Introduction header is bigger than the Introducer header in the other API intro pages.

Not sure why that would be. Interesting.

lucasmarshall:
> > Really minor - in Management API intro page, the Introduction header is bigger than the Introducer header in the other API intro pages.
> 
> Not sure why that would be. Interesting.

Figured it out

PR #1173, opened 2023-06-30
chore: refactor modals

asdfryan:
Fixes SUP1-428

Refactor modal code to re-use same component. This also includes some other misc cleanups.

Also swaps cancel + confirm buttons.

## Test Plan

Tested locally

image

## Deployment instructions


apps/mgmt-ui/src/components/ApplicationMenu/index.tsx
```diff
@@ -226,7 +226,6 @@ function UpdateApplicationMenuItem({
           <Button
             onClick={() => {
               onUpdateApplicationName(application.id, applicationName);
-              // TODO: error state
```
asdfryan, line 229: We handle error state now

tomkit (approved):
nit comment: use a more specific name for these dialogs since they're all referencing some resource

PR #1176, opened 2023-06-30
chore: new search attributes

asdfryan:
Fixes: SUP1-422

Add application env and object type as temporal search attributes.
Also data drive log interceptor. 

## Test Plan

Testing locally

## Deployment instructions


lucasmarshall (approved):

packages/core/temporal/index.ts
```diff
@@ -4,7 +4,9 @@ const { IndexedValueType } = temporal.api.enums.v1;
 
 export const TEMPORAL_CUSTOM_SEARCH_ATTRIBUTES = {
   SYNC_ID: 'SyncId',
+  OBJECT_TYPE: 'ObjectType',
```
albertyfwu, line 7: did we purposely leave out `object_name`?

asdfryan: Oh shoot I forgot about that.

We're at 10, so we'd have to kick out an attribute.

PR #1178, opened 2023-06-30
chore: embed link should redirect back to current page by default

asdfryan:


albertyfwu (approved):

tomkit:
Do we want developers to have it redirect to our management portal?

tomkit:
I would rather have it fail fast than silently (accidentally shipping it in their production app)

albertyfwu:
> I would rather have it fail fast than silently (accidentally shipping it in their production app)

Yea actually that's probably the right behavior

PR #1180, opened 2023-07-01
fix(mgmt-ui): Fix provider cards in mgmt-ui

asdfryan:
Fixes: SUP1-401

- remove alpha/beta labels
- MsDynamics365Sales -> Microsoft Dynamics 365 Sales
- Make card height responsive so we never hide the configure button
## Test Plan

Tested locally

https://www.loom.com/share/4b9d0d316add43f7869ca532a9724305

## Deployment instructions


apps/mgmt-ui/src/components/configuration/provider/ProviderCard.tsx
```diff
@@ -21,7 +21,7 @@ export default function ProviderCard({
   return (
     <Card
       classes={{
-        root: 'h-48 justify-between flex flex-col overflow-y-hidden',
+        root: 'min-h-48 justify-between flex flex-col overflow-y-hidden',
```
albertyfwu, line 24: why does this work but `h-48` doesn't work?

albertyfwu (approved):

apps/mgmt-ui/src/components/configuration/provider/ProviderCard.tsx
asdfryan: `h-48` makes it a fixed height while `min-h-48` allows it to grow if needed.

PR #1184, opened 2023-07-03
feat: rewrite custom field keys & values in both directions for pipedrive

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/remotes/impl/pipedrive/index.ts
+++ b/packages/core/remotes/impl/pipedrive/index.ts
@@ -50,6 +50,7 @@ import {
   toPipedrivePersonCreateParams,
   toPipedrivePersonUpdateParams,
 } from './mappers';
+import type { PipedriveObjectField } from './types';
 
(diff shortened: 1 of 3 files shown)


asdfryan (changes requested):
Where are we applying the schema fields for pipedrive? I think in order to support custom fields, we need to do that first, no?

albertyfwu:
> Where are we applying the schema fields for pipedrive? I think in order to support custom fields, we need to do that first, no?

yea didn't realize it's not implemented, though i would say it's orthogonal to this PR.

can we get this in first?

would prefer not to make massive PRs if not necessary

asdfryan (commented):
Is there any type coercion necessary? Does this work with non-string custom fields (i.e. integer or boolean)?

albertyfwu:
> Is there any type coercion necessary? Does this work with non-string custom fields (i.e. integer or boolean)?

not sure but i think that's not supaglue's responsibility

asdfryan (approved):

PR #1185, opened 2023-07-03
feat: API to trigger sync (with full refresh option)

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


apps/api/routes/mgmt/v2/sync/index.ts
```diff
@@ -0,0 +1,80 @@
...
+  const syncRouter = Router();
+
+  syncRouter.get(
+    '/',
+    async (
+      req: Request<GetSyncsPathParams, GetSyncsResponse, GetSyncsRequest, GetSyncsQueryParams>,
+      res: Response<GetSyncsResponse>
+    ) => {
+      function getObjectFilter() {
+        if (req.query?.object_type && req.query.object) {
```
asdfryan, line 28: IMO it's weird that the query params are `object` and `object_type` given that our representation of an `object` is a complex object with 2 fields -- feels like it should be `name` and `type`, or `object_name` and `object_type`.

apps/api/routes/mgmt/v2/sync/index.ts
```diff
@@ -0,0 +1,80 @@
...
+          return {};
+        }
+        throw new BadRequestError('object_type and object must both be present or both be absent');
+      }
+
+      const { next, previous, results, totalCount } = await objectSyncService.list({
+        applicationId: req.supaglueApplication.id,
+        paginationParams: toPaginationInternalParams({ page_size: req.query?.page_size, cursor: req.query?.cursor }),
+        ...getObjectFilter(),
+        externalCustomerId: req.query?.customer_id,
```
asdfryan, line 43: Shouldn't these also be required?

apps/api/routes/mgmt/v2/sync/index.ts
```diff
@@ -0,0 +1,80 @@
...
+        object_type: result.objectType,
+        object: result.object,
+        connection_id: result.connectionId,
+        sync_config_id: result.syncConfigId,
+      }));
+      return res.status(200).send({ next, previous, results: snakeCaseResults, total_count: totalCount });
+    }
+  );
+
+  syncRouter.post(
```
asdfryan, line 58: Why do you need to have the sync_id to trigger and not the object name, type, customer etc.

packages/core/mappers/object_sync.ts
```diff
@@ -0,0 +1,22 @@
...
+    connectionId: model.connectionId,
+    objectType: model.objectType as ObjectType,
+    object: model.object,
+    type,
+    syncConfigId: model.syncConfigId,
+    ...otherStrategyProps,
+    state: model.state as ObjectSyncState,
+    paused: model.paused,
+    argsForNextRun: model.argsForNextRun,
+  } as ObjectSync;
```
asdfryan, line 21: What happens if you don't do the type assertion here?

apps/api/services/connection_and_sync_service.ts
```diff
@@ -47,6 +50,45 @@ export class ConnectionAndSyncService {
...
+        argsForNextRun: {
+          performFullRefresh,
+        },
+      },
+    });
+
+    // Trigger a run in Temporal
+    const handle = this.#temporalClient.schedule.getHandle(getRunObjectSyncScheduleId(objectSyncId));
+    // TODO: is this the right policy? should it be configurable?
+    await handle.trigger(ScheduleOverlapPolicy.CANCEL_OTHER);
```
asdfryan, line 87: Shouldn't we also update sync logs? Right now it might leave it in an awkward started but never finished state.

asdfryan: Also maybe emit webhooks

apps/api/routes/mgmt/v2/sync/index.ts
albertyfwu: our convention everywhere else in documentation is that `object` refers to `account`, `contact`, etc. shouldn't we use this convention? are you suggesting to call it `object_name` everywhere or just here, while leaving other uses of `object` untouched? @george-xing thoughts?

apps/api/routes/mgmt/v2/sync/index.ts
albertyfwu: why do these need to be required? this is the same filter we are using for sync history / sync runs as well

apps/api/routes/mgmt/v2/sync/index.ts
albertyfwu: this seems more RESTful. for example, we also have URLs that have `/connections/:connection_id` and we do not replace those endpoints with calling `/connections` with body params to filter for a connection instead

if we replace everything with object name, type, customer, provider_name, application_id, i think we should adopt that convention everywhere. thoughts? @tomkit @george-xing

packages/core/mappers/object_sync.ts
albertyfwu: type issue because `ObjectSync` is a union type. In order to avoid this issue, we would need to do something like

```
if (type === 'full then incremental') {
  return {
    ...base,
    // more stuff
  };
} else {
  return {
    ...base,
    // more stuff
  }
}
```

apps/api/services/connection_and_sync_service.ts
albertyfwu: why do we need to update sync logs here? syncs logs are created inside the temporal workflow itself. same with webhooks.

apps/api/routes/mgmt/v2/sync/index.ts
asdfryan: Then why are `object` and `object_type` required?

apps/api/services/connection_and_sync_service.ts
asdfryan: The problem is you are potentially stopping a sync midway and not emitting a webhook or updating sync logs. We need to update those so that they are no longer "in progress"

apps/api/routes/mgmt/v2/sync/index.ts
asdfryan: Sure, that's fine.

apps/api/routes/mgmt/v2/sync/index.ts
albertyfwu: they are not required. they should either be both present or both not be present, since they don't make much sense in isolation

apps/api/services/connection_and_sync_service.ts
albertyfwu: i purposely selected `CANCEL_OTHER` as the strategy and not `TERMINATE_OTHER`, which would not allow an existing workflow to clean up. if our workflow handles things properly (try / catch), the cancellation exception should be caught and exited gracefully

albertyfwu:
@asdfryan can you re-review?

asdfryan (approved):

PR #1197, opened 2023-07-06
feat: custom objects and records API for salesforce

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


openapi/v2/crm/components/schemas/create_update_custom_object.yaml
```diff
@@ -19,6 +19,9 @@ properties:
     required:
       - singular
       - plural
+  primary_field_key_name:
```
albertyfwu, line 22: both salesforce & hubspot require this. if future providers don't require it, we'll have to figure out what to do with this

packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -447,6 +458,367 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
...
+            label: field.displayName,
+            type: field.fieldType === 'string' ? 'Text' : 'Number',
+            required: field.isRequired,
+          };
+
+          if (field.fieldType === 'string') {
+            return {
+              ...base,
+              // TODO: support configurations per field type
+              length: 255,
```
albertyfwu, line 538: if we want to not use defaults here, we'd need to expand our API

packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -447,6 +458,367 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
...
+            };
+          } else {
+            return {
+              ...base,
+              // TODO: support configurations per field type
+              precision: field.fieldType === 'number' ? 18 : undefined,
```
albertyfwu, line 544: if we want to not use defaults here, we'd need to expand our API

packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -447,6 +458,348 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
...
+    const metadata = await this.#client.metadata.read('CustomObject', id);
+    return toCustomObject(metadata);
+  }
+
+  public override async createCustomObject(params: CustomObjectCreateParams): Promise<string> {
+    if (!params.name.endsWith('__c')) {
+      // Salesforce doesn't actually enforce this, and will just append __c.
+      // However, we want to enforce this to avoid confusion when using
+      // the custom object name in other places.
+      throw new Error('Custom object name must end with __c');
```
asdfryan, line 478: This should be a 4xx and not a 5xx right?

packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -447,6 +458,348 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
...
+            required: field.isRequired,
+          };
+
+          if (field.fieldType === 'string') {
+            return {
+              ...base,
+              // TODO: support configurations per field type
+              length: 255,
+            };
+          } else {
```
asdfryan, line 540: Nit: else is redundant

openapi/v2/crm/components/schemas/create_update_custom_object.yaml
asdfryan: Can you add a description?

packages/core/remotes/impl/salesforce/index.ts
asdfryan: We should document this in the API spec

packages/core/remotes/impl/salesforce/index.ts
asdfryan: We should document this in the API spec

packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -447,6 +458,372 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
...
+      throw new Error('Custom object id must end with __c');
+    }
+
+    const metadata = await this.#client.metadata.read('CustomObject', id);
+    return toCustomObject(metadata);
+  }
+
+  public override async createCustomObject(params: CustomObjectCreateParams): Promise<string> {
+    if (!params.name.endsWith('__c')) {
+      // Salesforce doesn't actually enforce this, and will just append __c.
```
tomkit, line 475: Are these documented in openapi?

albertyfwu: no i'll add it

PR #1202, opened 2023-07-08
chore: support bigquery as a sync destination

lucasmarshall:
## Test Plan

Tested locally against our bigquery/hubspot/salesforce


packages/core/destination_writers/bigquery.ts
lucasmarshall, line 1: This is the main file for this, so be sure to expand it and review

packages/core/destination_writers/bigquery.ts
```diff
@@ -0,0 +1,1389 @@
...
+        c !== 'id'
+    );
+
+    // Output
+    const stream = client
+      .dataset(dataset)
+      .table(tempTable)
+      .createWriteStream({
+        sourceFormat: 'CSV',
+        createDisposition: 'CREATE_IF_NEEDED',
```
albertyfwu, line 105: what does this setting do? does this create the temp table if it doesn't exist? if so, does this obviate the need for `await client.dataset(dataset).createTable(table, { schema: getCommonObjectSchema(category, commonObjectType) });` ?

packages/core/destination_writers/bigquery.ts
```diff
@@ -0,0 +1,1389 @@
...
+      stringifier,
+      stream
+    );
+    childLogger.info({ tempTable }, 'Importing common object objects into temp table [COMPLETED]');
+
+    heartbeat();
+
+    // Copy from deduped temp table
+    childLogger.info({ table, tempTable }, 'Copying from deduped temp table to main table [IN PROGRESS]');
+    // IMPORTANT: we need to use DISTINCT ON because we may have multiple records with the same id
```
albertyfwu, line 167: nit: adjust comment text for bigquery

packages/core/destination_writers/bigquery.ts
```diff
@@ -0,0 +1,1389 @@
...
+    );
+    const columnsWithLastModifiedAt = [...columns, '_supaglue_last_modified_at'];
+
+    // Output
+    const stream = client
+      .dataset(dataset)
+      .table(tempTable)
+      .createWriteStream({
+        sourceFormat: 'CSV',
+        createDisposition: 'CREATE_IF_NEEDED',
```
albertyfwu, line 255: same question here

packages/core/destination_writers/bigquery.ts
```diff
@@ -0,0 +1,1389 @@
...
+      stringifier,
+      stream
+    );
+    childLogger.info({ tempTable }, 'Importing raw records into temp table [COMPLETED]');
+
+    heartbeat();
+
+    // Copy from deduped temp table
+    childLogger.info({ table, tempTable }, 'Copying from deduped temp table to main table [IN PROGRESS]');
+    // IMPORTANT: we need to use DISTINCT ON because we may have multiple records with the same id
```
albertyfwu, line 318: same nit about comment text

albertyfwu (approved):
lgtm. left some non-blocking comments.

packages/core/destination_writers/bigquery.ts
lucasmarshall: That line doesn't create the temp table, but it does create the destination table.

packages/core/destination_writers/bigquery.ts
lucasmarshall: Good catch

packages/core/mappers/destination.ts
```diff
@@ -1,12 +1,14 @@
...
 
 export const fromDestinationModel = (model: DestinationModel): Destination => {
+  const config = model.config as any;
   return {
     id: model.id,
     name: model.name,
     type: model.type,
     applicationId: model.applicationId,
-    config: model.config,
+    config: model.type === 'bigquery' ? { ...config, credentials: camelcaseKeys(config.credentials) } : config,
```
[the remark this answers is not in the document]
lucasmarshall: It's mostly so I don't have to hack the type generation stuff to generate snake case keys in this case only. This was the easiest way to handle it that I could find.

openapi/v2/mgmt/components/schemas/create_update_destination.yaml
```diff
@@ -7,27 +7,39 @@ allOf:
...
+            type: string
+            enum:
+              - postgres
+          config:
+            $ref: ./objects/postgres_config.yaml
+        required:
+          - type
+          - config
+      - type: object
+        title: postgres
```
tomkit, line 35: bigquery?

lucasmarshall: Good catch. Will fix.

packages/core/destination_writers/bigquery.ts
```diff
@@ -0,0 +1,1391 @@
...
+    const { dataset } = this.#destination.config;
+    const table = getCommonObjectTableName(category, commonObjectType);
+    const qualifiedTable = `${dataset}.${table}`;
+    const tempTable = `_temp_${providerName}_${table}`;
+    const qualifiedTempTable = `${dataset}.${tempTable}`;
+
+    const client = await this.#getClient();
+
+    // Create table if necessary
+    // TODO: We should only need to do this once at the beginning
```
tomkit, line 79: Would this work if we evolve our common object schema?

lucasmarshall: We'd have to change the schema here too.

packages/core/destination_writers/bigquery.ts
```diff
@@ -0,0 +1,1391 @@
...
+
+    const { dataset } = this.#destination.config;
+    const table = getObjectTableName(providerName, object);
+    const qualifiedTable = `${dataset}.${table}`;
+    const tempTable = `_temp_${providerName}_${table}`;
+    const qualifiedTempTable = `${dataset}.${tempTable}`;
+
+    const client = await this.#getClient();
+
+    // Create table if necessary
```
tomkit, line 219: Same comment as above: this will cause issues if we evolve our schema and there is a dangling temp table (?)

tomkit:
Can you add a BigQuery destinations documentation page?

packages/core/destination_writers/bigquery.ts
```diff
@@ -0,0 +1,1391 @@
...
+        c !== '_supaglue_application_id' &&
+        c !== '_supaglue_provider_name' &&
+        c !== '_supaglue_customer_id' &&
+        c !== 'id'
+    );
+
+    // Output
+    const stream = client
+      .dataset(dataset)
+      .table(tempTable)
```
tomkit, line 102: Do we need to potentially remove dangling temp tables?

lucasmarshall: We do below.

PR #1219, opened 2023-07-11
chore: make mapped field optional when updating

asdfryan:


packages/core/services/connection_service.ts
```diff
@@ -188,7 +188,7 @@ export class ConnectionService {
       .map(({ name }) => name);
     for (const requiredField of schemaMappingRequiredFields) {
       if (!params.fieldMappings.find(({ schemaField }) => schemaField === requiredField)) {
-        throw new BadRequestError(`Field ${requiredField} is required to be set by the schema`);
+        throw new BadRequestError(`Field ${requiredField} is missing.`);
```
asdfryan, line 191: We require all fields to be present, but the mappedField can be undefined (to prevent error)

albertyfwu (approved):

PR #1222, opened 2023-07-11
chore: fix unique constraint on object sync

asdfryan:


lucasmarshall (approved):

packages/db/prisma/schema.prisma
```diff
@@ -140,7 +140,7 @@ model ObjectSync {
   updatedAt      DateTime        @updatedAt @map("updated_at")
   runs           ObjectSyncRun[]
 
-  @@unique([connectionId, object])
+  @@unique([connectionId, object, objectType])
```
albertyfwu, line 143: might be better to have `objectType` before `object`, but this works

albertyfwu (approved):

PR #1233, opened 2023-07-13
feat: sync standard objects for gong

albertyfwu:
Supports syncing standard objects for Gong.

Caveats:
- even if a field mapping is set, it's ignored
- haven't tested real rate limiting (429), because I didn't run into it while testing; we can artificially try to hit 429 if we want

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/services/sync_config_service.ts
+++ b/packages/core/services/sync_config_service.ts
@@ -70,10 +70,21 @@ export class SyncConfigService {
   public async create(syncConfig: SyncConfigCreateParams): Promise<SyncConfig> {
     validateSyncConfigParams(syncConfig);
     // TODO:(SUP1-350): Backfill sync schedules for connections
-    const createdSyncConfig = await this.#prisma.syncConfig.create({
-      data: toSyncConfigModel(syncConfig),
+    const createdSyncConfigModel = await this.#prisma.$transaction(async (tx) => {
+      const createdSyncConfigModel = await tx.syncConfig.create({
+        data: toSyncConfigModel(syncConfig),
+      });
+
+      await tx.syncConfigChange.create({
+        data: {
+          syncConfigId: createdSyncConfigModel.id,
+        },
+      });
+
+      return createdSyncConfigModel;
     });
-    return fromSyncConfigModel(createdSyncConfig);
+
+    return fromSyncConfigModel(createdSyncConfigModel);
   }
 
   public async update(id: string, applicationId: string, params: SyncConfigUpdateParams): Promise<SyncConfig> {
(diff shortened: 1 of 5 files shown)


packages/core/services/sync_config_service.ts
```diff
@@ -70,10 +70,21 @@ export class SyncConfigService {
...
     validateSyncConfigParams(syncConfig);
     // TODO:(SUP1-350): Backfill sync schedules for connections
-    const createdSyncConfig = await this.#prisma.syncConfig.create({
-      data: toSyncConfigModel(syncConfig),
+    const createdSyncConfigModel = await this.#prisma.$transaction(async (tx) => {
+      const createdSyncConfigModel = await tx.syncConfig.create({
+        data: toSyncConfigModel(syncConfig),
+      });
+
+      await tx.syncConfigChange.create({
```
albertyfwu, line 78: this is a bug where we weren't creating ObjectSyncs when SyncConfig was updated with new objects.

packages/core/services/sync_config_service.ts
```diff
@@ -103,14 +114,25 @@ export class SyncConfigService {
...
   public async upsert(syncConfig: SyncConfigCreateParams): Promise<SyncConfig> {
     validateSyncConfigParams(syncConfig);
-    const upsertedSyncConfig = await this.#prisma.syncConfig.upsert({
-      where: {
-        providerId: syncConfig.providerId,
-      },
-      create: toSyncConfigModel(syncConfig),
-      update: toSyncConfigModel(syncConfig),
+
+    const upsertedSyncConfigModel = await this.#prisma.$transaction(async (tx) => {
```
albertyfwu, line 118: another bug

packages/core/services/sync_config_service.ts
```diff
@@ -122,9 +144,16 @@ export class SyncConfigService {
     if (objectSyncs.length) {
       throw new BadRequestError('Cannot delete sync config with active connections');
     }
-    await this.#prisma.syncConfig.deleteMany({
-      where: { id, applicationId },
-    });
+    await this.#prisma.$transaction([
```
albertyfwu, line 147: another bug

tomkit:
some thoughts:
- can we verify the user wants to consume these as three separate tables?
- can we validate ui and api input to only accept the known enums and/or provide a select option in the mgmt-ui the way we do for salesforce
- thoughts on just using lowercased names? hubspot is lower case, salesforce is pascal. we don't need to add camelCasing if it's our own choosing

PR #1238, opened 2023-07-14
fix: common object syncconfig

albertyfwu:
- show correct common object selections for provider category
- don't autofill by default

## Test Plan

[Describe test plan here]

## Deployment instructions


tomkit (approved):
ty

PR #1240, opened 2023-07-14
feat: allow trigger of sync on mgmt-ui

albertyfwu:
no that pretty but works for now.

later, we probably want to show this on the Sync Runs page also.
Also, we shouldn't show the button if a sync is already running, for better UX.

image

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/mgmt-ui/src/components/logs/SyncsTable.tsx
+++ b/apps/mgmt-ui/src/components/logs/SyncsTable.tsx
@@ -1,6 +1,7 @@
-import { pauseSync, resumeSync } from '@/client';
+import { pauseSync, resumeSync, triggerSync } from '@/client';
 import { useNotification } from '@/context/notification';
 import { useActiveApplicationId } from '@/hooks/useActiveApplicationId';
+import { Button } from '@mui/material';
 import Switch from '@mui/material/Switch';
 import type { GridColDef } from '@mui/x-data-grid';
 import { DataGrid } from '@mui/x-data-grid';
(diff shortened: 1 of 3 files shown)


lucasmarshall (approved):

apps/mgmt-ui/src/components/logs/SyncsTable.tsx
```diff
@@ -99,6 +100,36 @@ export default function SyncsTable(props: SyncsTableProps) {
...
+    {
+      field: '_',
+      headerName: 'Actions',
+      width: 120,
+      sortable: false,
+      renderCell: (params) => {
+        // button to trigger sync run
+        return (
+          <Button
+            variant="contained"
```
tomkit, line 112: nit: `size="small"`

tomkit: nit: https://mui.com/material-ui/react-tooltip/ to help message to the user the impact of a full refresh

tomkit (approved):
ty

apps/mgmt-ui/src/components/logs/SyncsTable.tsx
albertyfwu: there's no full refresh here. we can do that later.

tomkit: sg

thoughts on quick-fix? would adding two buttons to the same row allow the user to do either): `trigger` and `trigger full`?

albertyfwu: ok

PR #1245, opened 2023-07-17
feat: embed link dropdown

albertyfwu:
Don't make developer replace `{{REPLACE_ME}}` and allow them to select from a dropdown of providers instead.

https://www.loom.com/share/dcf3f8d9e1eb45098510f1a61ddbf90d

## Test Plan

[Describe test plan here]

## Deployment instructions


tomkit (approved):
🙏

PR #1252, opened 2023-07-18
feat: Add support for Apollo provider and connection

asdfryan:
Fixes: SUP1-461, SUP1-464

This PR:
- Adds support for API-key based connections (previously only supported Oauth)
- Adds a POST endpoint for connections only for API key based connections (just Apollo at the moment)
- Adds Apollo provider to mgmt UI

## Test Plan

Tested locally

## Deployment instructions


packages/core/remotes/impl/apollo/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    throw new Error('Not implemented');
+  }
+}
+
+export function newClient(connection: ConnectionUnsafe<'apollo'>, provider: Provider): ApolloClient {
+  return new ApolloClient();
+}
+
+export const authConfig: ConnectorAuthConfig = {
+  tokenHost: '',
```
tomkit, line 20: nit: leave comment about api key only?

tomkit (approved):

packages/core/remotes/impl/apollo/index.ts
asdfryan: Sure, in next PR.

PR #1253, opened 2023-07-18
feat: support apollo passthrough

asdfryan:
Fixes: SUP1-465

It seems the Apollo API uses the body for the api token for POST/PUT/PATCH while using the query parameters for GET requests, so we need to override the `sendPassthroughRequest` to properly do this.

## Test Plan

Tested locally

## Deployment instructions

### DIFF
--- a/packages/core/remotes/impl/apollo/index.ts
+++ b/packages/core/remotes/impl/apollo/index.ts
@@ -1,19 +1,43 @@
-import type { ConnectionUnsafe, Provider } from '@supaglue/types';
+import type {
+  ConnectionUnsafe,
+  Provider,
+  SendPassthroughRequestRequest,
+  SendPassthroughRequestResponse,
+} from '@supaglue/types';
 import type { ConnectorAuthConfig } from '../../base';
 import { AbstractEngagementRemoteClient } from '../../categories/engagement/base';
 
 class ApolloClient extends AbstractEngagementRemoteClient {
-  public constructor() {
-    super('missing-base-url');
+  #apiKey: string;
+
+  public constructor(apiKey: string) {
+    super('https://api.apollo.io');
+    this.#apiKey = apiKey;
   }
 
   protected override getAuthHeadersForPassthroughRequest(): Record<string, string> {
-    throw new Error('Not implemented');
+    return {};
+  }
+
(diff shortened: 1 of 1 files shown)


packages/core/remotes/impl/apollo/index.ts
```diff
@@ -1,19 +1,43 @@
...
   protected override getAuthHeadersForPassthroughRequest(): Record<string, string> {
-    throw new Error('Not implemented');
+    return {};
+  }
+
+  public override async sendPassthroughRequest(
+    request: SendPassthroughRequestRequest
+  ): Promise<SendPassthroughRequestResponse> {
+    // Apollo requires the api key to be in the query string for GET requests, and in the body for POST/PATCH/PUT.
+    // Note: This is not explicitly documented, but seems to be the convention for their API
```
tomkit, line 26: HTTP GETs don't have a body so it's not an API convention thing, but related to the HTTP protocol

tomkit (approved):

packages/core/remotes/impl/apollo/index.ts
```diff
@@ -1,19 +1,43 @@
...
+    if (request.method === 'GET') {
+      return await super.sendPassthroughRequest({ ...request, query: { ...request.query, api_key: this.#apiKey } });
+    }
+    if (request.method === 'POST' || request.method === 'PATCH' || request.method === 'PUT') {
+      const bodyJson = request.body ? JSON.parse(request.body as string) : {};
```
albertyfwu, line 31: how do we know that the body has to be JSON? do we want to make that assumption for all providers? If we're going to assume that the body is JSON, we should just change the openAPI spec to say this is object, and not string.

albertyfwu (approved):

packages/core/remotes/impl/apollo/index.ts
asdfryan: Yes I know this.

But it can still be in the query string for POST/PATCH/PUT. Apollo has made the explicit choice to put it in the body for those requests.

packages/core/remotes/impl/apollo/index.ts
albertyfwu: can you make a task for that?

PR #1254, opened 2023-07-18
fix: issues when doing mongo runthrough

lucasmarshall:
Ran into a few issues that weren't noticed before during previous runthroughs.

## Test Plan

Tested locally against mongodb atlas

### DIFF
--- a/packages/core/destination_writers/mongodb.ts
+++ b/packages/core/destination_writers/mongodb.ts
@@ -69,7 +69,7 @@ export class MongoDBDestinationWriter extends BaseDestinationWriter {
 
     await collection.replaceOne(
       {
-        _id: mappedRecord.id,
+        id: mappedRecord.id,
         _supaglue_application_id: mappedRecord._supaglue_application_id,
         _supaglue_provider_name: mappedRecord._supaglue_provider_name,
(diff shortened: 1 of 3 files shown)


packages/core/mappers/destination.ts
```diff
@@ -89,6 +91,18 @@ export const fromDestinationModelToSafe = (model: DestinationModel): Destination
         },
       };
     }
+    case 'mongodb': {
```
albertyfwu, line 94: oops my bad

tomkit (approved):

packages/core/destination_writers/mongodb.ts
```diff
@@ -279,7 +279,7 @@ function streamToMongoDBCollection(collection: Collection, client: MongoClient,
       documents.map((document) => ({
         replaceOne: {
           filter: {
-            _id: document.id,
+            _id: document.id ?? document._id,
```
albertyfwu, line 282: in what scenario do we have `_id` or `id`? do we set this when pulling from provider?

lucasmarshall: I've clarified this by switching to `id` for our own ids. `_id` will be autogenerated by mongo.

PR #1260, opened 2023-07-19
chore: update field mapping openapi spec to not use oneOf/anyOf

asdfryan:
This makes it so that the nested `fields` object spec gets expanded

Screenshot 2023-07-19 at 10 51 18 AM


albertyfwu (commented):
can you explain why we need this change?

asdfryan:
> can you explain why we need this change?

We need this because our current APISpec -> Doc tooling does not support anyOf and oneOf and it wasn't expanding the nested object.

asdfryan:
> You can still get the semantics right here and have it work correctly in docs. Just use a `oneOf` at the top level and duplicate.

That won't work because there is overlap (because both are optional). What we need is anyOf. 

In any case I personally think a top level branching would make it more confusing, and we can just document that only one should be set (or one should take precedence over another)

albertyfwu (approved):

PR #1265, opened 2023-07-19
feat: for mongodb, normalize fields

lucasmarshall:
## Test Plan

Tested locally against mongodb atlas and hubspot/salesforce


albertyfwu (approved):
ideally we should do this mapping higher up in the stack, and each destination writer shouldn't need to know to flatten hubspot entries, but this works for now.

when i'm done with the schema-centric/object-centric work, the interface on the provider clients to list objects may change a bit, so i'll try to streamline it a bit at that time.

PR #1270, opened 2023-07-20
feat: endpoint to get logged in user id for a connection by provider name

lucasmarshall:
Fixes: SUP1-477 

## Test Plan

Tested locally against dynamics


apps/api/routes/mgmt/v2/customer/connection/index.ts
```diff
@@ -35,6 +39,30 @@ export default function init(app: Router): void {
...
+    '/_provider_user_id',
+    async (
+      req: Request<
+        GetProviderUserIdPathParams,
+        GetProviderUserIdResponse,
+        GetProviderUserIdRequest,
+        GetProvideruserIdQueryParams
+      >,
+      res: Response<GetProviderUserIdResponse>
+    ) => {
```
asdfryan, line 52: Should we throw a bad request error if the provider is not msft365?

openapi/v2/mgmt/paths/connections@_provider_user_id.yaml
```diff
@@ -0,0 +1,36 @@
...
+  tags:
+    - Connections
+  security:
+    - ApiKeyAuth: []
+  parameters:
+    - name: provider_name
+      in: query
+      required: true
+      schema:
+        type: string
```
asdfryan, line 13: We have a provider_name enum. (provider_name.yaml)

apps/api/routes/mgmt/v2/customer/connection/index.ts
lucasmarshall: Sure

asdfryan (approved):

PR #1272, opened 2023-07-20
fix: return a 502 if dynamics api is unreachable

lucasmarshall:


packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -1338,7 +1338,7 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
       case 'ERROR_HTTP_304':
         return new NotModifiedError(error.message, error);
       case 'ERROR_HTTP_503':
-        return new ServiceUnavailableError('Salesforce API is temporarily unavailable', error);
+        return new BadGatewayError('Salesforce API is temporarily unavailable', error);
```
lucasmarshall, line 1341: This was the wrong error code to be returning in this case.

albertyfwu (approved):

PR #1275, opened 2023-07-21
feat: salesloft provider + connection

asdfryan:
Note: I'm making declaring scopes disabled (it should just take the scopes of the developer app) since it doesn't seem to be supported / I can't figure out what the right names are.


apps/api/routes/oauth.ts
```diff
@@ -109,13 +109,13 @@ export default function init(app: Router): void {
 
       const authorizationUri = client.authorizeURL({
         redirect_uri: REDIRECT_URI,
-        scope: oauthScopes,
+        scope: provider.name === 'salesloft' ? oauthScopes : [],
```
lucasmarshall, line 112: This doesn't seem correct...

lucasmarshall (changes requested):

lucasmarshall (approved):

PR #1288, opened 2023-07-24
chore: return all entities in /entity_mappings

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


packages/types/entity_mapping.ts
```diff
@@ -1,13 +1,17 @@
...
 
-type From = {
-  from: 'developer' | 'customer';
-};
-
 export type MergedEntityMapping = {
   entityId: string;
-  object?: StandardOrCustomObject & From;
-  fieldMappings?: (EntityFieldMapping & From)[];
+  object?: StandardOrCustomObject & {
```
asdfryan, line 5: Could you also add the `entityName` here?

albertyfwu: ok

asdfryan (approved):

packages/types/entity_mapping.ts
```diff
@@ -1,13 +1,18 @@
...
-
 export type MergedEntityMapping = {
   entityId: string;
-  object?: StandardOrCustomObject & From;
-  fieldMappings?: (EntityFieldMapping & From)[];
+  entityName: string;
+  object?: StandardOrCustomObject & {
+    from: 'developer' | 'customer';
+  };
+  fieldMappings: MergedEntityFieldMapping[];
```
asdfryan, line 9: Shouldn't this be optional? It's optional in the openAPI spec

PR #1293, opened 2023-07-24
chore: list standard objects for sf and hs for UI

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -245,6 +247,13 @@ class SalesforceClient extends AbstractCrmRemoteClient {
     });
   }
 
+  public override async listObjects(): Promise<StandardOrCustomObject[]> {
+    return SALESFORCE_OBJECTS.map((object) => ({
```
asdfryan, line 252: We need to actually query Salesforce I think

albertyfwu: ok

asdfryan (approved):

albertyfwu:
let's get back to this when we actually pull data from salesforce and hubspot. right now this is useless

PR #1300, opened 2023-07-24
feat: support adding additional field mappings

asdfryan:
Screenshot 2023-07-24 at 4 51 47 PM


tomkit (approved):

packages/types/entity_mapping.ts
```diff
@@ -13,6 +14,7 @@ export type MergedEntityFieldMapping = {
   entityField: string;
   mappedField?: string;
   from?: 'developer' | 'customer';
+  isAdditional: boolean;
```
albertyfwu, line 17: technically this should can only be `true` if `from`: is `customer`, so the type can be stricter, but this is OK for now

albertyfwu (approved):

PR #1315, opened 2023-07-25
chore: write mapped and raw data to separate columns

albertyfwu:
tested for salesforce -> postgres and hubspot -> mongo

## Test Plan

[Describe test plan here]

## Deployment instructions


packages/core/destination_writers/mongodb.ts
```diff
@@ -249,6 +249,7 @@ export class MongoDBDestinationWriter extends BaseDestinationWriter {
               _supaglue_emitted_at: record.emittedAt,
               _supaglue_is_deleted: record.isDeleted,
               _supaglue_raw_data: record.rawData,
+              _supaglue_mapped_data: record.mappedData,
```
lucasmarshall, line 252: We also need to switch the normalized data to using this, right?

albertyfwu: yea you're right

packages/core/destination_writers/mongodb.ts
```diff
@@ -229,13 +229,15 @@ export class MongoDBDestinationWriter extends BaseDestinationWriter {
       inputStream,
       new Transform({
         objectMode: true,
-        transform: (record: NormalizedRawRecord, encoding, callback) => {
+        transform: (record: ObjectRecord, encoding, callback) => {
           // normalized fields are the same as raw fields for most providers, at least for now
           // we really want this to only be the mapped fields
```
lucasmarshall, line 234:
```suggestion
```

lucasmarshall (approved):
small nit, otherwise looks good

PR #1319, opened 2023-07-26
feat: support gong connection via access key + secret

asdfryan:
Tested locally


packages/core/remotes/impl/gong/index.ts
```diff
@@ -46,21 +49,41 @@ type GongClientConfig = {
...
+  accessKey: string;
+  accessKeySecret: string;
+};
+
 class GongClient extends AbstractEngagementRemoteClient {
   readonly #config: GongClientConfig;
 
   public constructor(config: GongClientConfig) {
-    super(config.instanceUrl);
+    super(config.type === 'oauth2' ? config.instanceUrl : 'https://api.gong.io');
```
lucasmarshall, line 62: Not sure we need this. According to https://app.gong.io/settings/api/documentation#overview :

> Base URL for all API calls is https://api.gong.io/

asdfryan: I am preserving what the behavior here was before @albertyfwu .

Also if you are logged into your developer account (that you'd use for OAuth), you see something different:
image

lucasmarshall (approved):

PR #1335, opened 2023-07-28
chore: do not allow dashes in entity names

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


asdfryan (approved):

asdfryan (commented):
Can we validate on the FE as well?

albertyfwu:
> Can we validate on the FE as well?

https://github.com/supaglue-labs/supaglue/pull/1336

PR #1349, opened 2023-07-31
feat: use svix for all webhooks

lucasmarshall:
## Test Plan

Tested locally with svix play


openapi/v2/mgmt/openapi.yaml
```diff
@@ -234,6 +246,17 @@ components:
       $ref: ../../common/components/parameters/header/x-provider-name.yaml
     x-customer-id:
       $ref: ../../common/components/parameters/header/x-customer-id.yaml
-webhooks:
-  webhook:
-    $ref: ./webhooks/webhook.yaml
+x-webhooks: &x-webhooks
```
lucasmarshall, line 249: Some tooling expects `x-webhooks`, while other tooling expects `webhooks`

openapi/versioned/version-0.12.10/common/components/parameters/header/x-customer-id.yaml
lucasmarshall, line 1: We no longer use versioned openapi specs

apps/api/services/connection_and_sync_service.ts
```diff
@@ -369,6 +372,15 @@ export class ConnectionAndSyncService {
       errored = true;
       throw e;
     } finally {
+      const { credentials: _, ...paramsWithoutCredentials } = params;
+      await this.#webhookService.sendMessage(
```
albertyfwu, line 376: how important it is to have guaranteed webhooks? if it's critical, it's still possible for this webhook to not fire, since we can crash before this point, or this call fails on first try

openapi/v2/mgmt/openapi.yaml
albertyfwu: can you include this explanation as a comment in the code?

openapi/v2/mgmt/webhooks/salesforce_cdc.create.yaml
```diff
@@ -0,0 +1,41 @@
...
+          fields:
+            Name: 'Acme'
+        schema:
+          type: object
+          properties:
+            id:
+              type: string
+              description: The Salesforce ID of the record that was created
+              example: '0011t00000B0G6uAAF'
+            entityName:
```
albertyfwu, line 26: should we use snakecase to be consistent, or is this intentional to pass along CDC events as-is?

openapi/v2/mgmt/webhooks/salesforce_cdc.delete.yaml
```diff
@@ -0,0 +1,34 @@
...
+          id: '0011t00000B0G6uAAF'
+          entityName: Account
+        schema:
+          type: object
+          properties:
+            id:
+              type: string
+              description: The Salesforce ID of the record that was deleted
+              example: '0011t00000B0G6uAAF'
+            entityName:
```
albertyfwu, line 24: should we use snakecase to be consistent?

openapi/v2/mgmt/webhooks/salesforce_cdc.undelete.yaml
```diff
@@ -0,0 +1,41 @@
...
+          fields:
+            Name: 'Acme'
+        schema:
+          type: object
+          properties:
+            id:
+              type: string
+              description: The Salesforce ID of the record that was undeleted
+              example: '0011t00000B0G6uAAF'
+            entityName:
```
albertyfwu, line 26: should we use snakecase to be consistent?

openapi/v2/mgmt/webhooks/salesforce_cdc.update.yaml
```diff
@@ -0,0 +1,64 @@
...
+          changedFields: [Name]
+          diffFields: []
+        schema:
+          type: object
+          properties:
+            id:
+              type: string
+              description: The Salesforce ID of the record that was updated
+              example: '0011t00000B0G6uAAF'
+            entityName:
```
albertyfwu, line 29: should we use snakecase to be consistent?

openapi/v2/mgmt/webhooks/salesforce_cdc.update.yaml
```diff
@@ -0,0 +1,64 @@
...
+              type: string
+              description: The name of the Salesforce object that the record belongs to
+              example: Account
+            nulledFields:
```
albertyfwu, line 33: should we use snakecase to be consistent?

openapi/versioned/version-0.12.10/common/components/parameters/header/x-customer-id.yaml
albertyfwu: why is that?

albertyfwu (approved):
some comments, but i don't think they're blocking

apps/api/services/connection_and_sync_service.ts
lucasmarshall: That is true. This makes it as durable as we can get without having a job queue, which I think adds too much complexity for now.

lucasmarshall: Thoughts here @george-xing @tomkit ?

openapi/versioned/version-0.12.10/common/components/parameters/header/x-customer-id.yaml
lucasmarshall: We only used it for docs and the generated docs themselves are now versioned.

openapi/v2/mgmt/webhooks/salesforce_cdc.create.yaml
lucasmarshall: We can. The original intent was for it to be consistent with the CDC events, but you are right that since our other webhooks use snake case we probably should here

albertyfwu: don't feel strongly here; if it's easier to keep shape as-is from CDC, then that's fine too

apps/api/services/connection_and_sync_service.ts
albertyfwu: regardless, if that's needed, we could tackle that in another PR, since this one is already big and isn't a regression

tomkit: They're using it as a notification to trigger an incremental ETL task twice a day. Since it's an infrequent task, they want to retry sooner than in 12 hours if, for some reason the webhook failed. 

Things will self-resolve upon the next webhook trigger for our customers, so we don't need to need at-least-once/exactly-once semantics here

PR #1364, opened 2023-08-02
feat: linear.app auth, standard object syncs

tomkit:
## Test Plan

[Describe test plan here]

## Deployment instructions


docs/docs/providers/linear.mdx
```diff
@@ -0,0 +1,22 @@
...
+| ----------------------- | --------- |
+| Authentication (Oauth2) | Yes       |
+| Managed syncs           | Yes       |
+| Actions API             | No        |
+| Real-time events        | No        |
+| Passthrough API         | Yes       |
+
+Supported Standard Objects: `issues`, `comments`, `teams`, `projects`, `users`.
+
+Scopes: `read`, `write`, `issues:create`, `comments:create`. `admin` for full privileges.
```
lucasmarshall, line 22: This is a little confusing. Some people may read that `admin` is required, but it shouldn't be, right?

packages/core/remotes/impl/linear/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  clientId: string;
+  clientSecret: string;
+};
+
+class LinearClient extends AbstractNoCategoryRemoteClient {
+  readonly #client: NativeLinearClient;
+  readonly #config: LinearClientConfig;
+  public constructor(config: LinearClientConfig) {
+    super('https://api.linear.app/graphql');
+    const { accessToken } = config;
```
lucasmarshall, line 29: Does this expire and need regular renewal? If so, can we implement that like we do in the other remotes?

lucasmarshall (changes requested):

packages/core/remotes/impl/linear/index.ts
lucasmarshall: BTW, my preferred implementation of this is in the dynamics remote

tomkit: I took a look: they don't have token refreshing. They just set a really long expiry (10 yrs)

lucasmarshall (approved):

PR #1366, opened 2023-08-03
fix: convert opportunity.amount (common schema) to float

tomkit:
- remove trunc after we migrate customer tables

## Test Plan

[Describe test plan here]

## Deployment instructions


packages/core/destination_writers/postgres.ts
```diff
@@ -733,7 +733,7 @@ CREATE ${temp ? 'TEMP TABLE' : 'TABLE'} IF NOT EXISTS ${
   "last_modified_at" TIMESTAMP(3) NOT NULL,
   "name" TEXT,
   "description" TEXT,
-  "amount" INTEGER,
+  "amount" FLOAT,
```
asdfryan, line 736: How will we migrate existing customers? (i.e. warmly)

albertyfwu (approved):

PR #1372, opened 2023-08-04
feat: custom objects, entities, associations

albertyfwu:
Change list (all of these are supported for Salesforce and HubSpot):
- Stop supporting custom objects for SyncConfig (no more syncing custom objects - need to use Entities instead)
- Allow mapping Entities to custom objects (FE and BE)
- Move AssociationTypes and Custom Objects to new Metadata API
- Support listing custom and standard objects from providers in Metadata API
- Support creating AssociationTypes between Entities
- Support creating Associations between Entity Records
- Update docs to show Metadata API
- Make creation of AssociationTypes and Entities more unified so that we perform provider-specific mutations; for example, if you pass in a name like `related_opportunity` to create an AssociationType in Salesforce, we will append `__c` to it to make sure it doesn't get rejected by the API
- moved properties endpoint to Metadata API
- changed `name` to `suggestedName` or something for creation of AssociationTypes and Custom Objects to make it clear that Supaglue may append `__c` to the name to make it work for Salesforce.
- ...other changes
- 
image

## Test Plan

[Describe test plan here]

## Deployment instructions


asdfryan (approved):

openapi/v2/actions/components/schemas/objects/entity_or_object_record.yaml
```diff
@@ -0,0 +1,43 @@
+oneOf:
```
tomkit, line 1: Can you add one full example here

openapi/v2/actions/components/schemas/objects/entity_or_object.yaml
```diff
@@ -0,0 +1,34 @@
+oneOf:
```
tomkit, line 1: Can you add one full example here

openapi/v2/actions/components/schemas/objects/custom_object_record.yaml
```diff
@@ -0,0 +1,17 @@
+type: object
+properties:
+  id:
+    type: string
+  custom_object_id:
+    type: string
+  data:
+    type: object
+    properties:
```
tomkit, line 9: can you add an example here

openapi/v2/actions/paths/objects/custom@{object_id}@{record_id}.yaml
```diff
@@ -0,0 +1,72 @@
...
+  security:
+    - x-api-key: []
+  responses:
+    200:
+      description: Custom object record
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/objects/custom_object_record.yaml
+          examples:
```
tomkit, line 15: Can we omit `examples: Example: value:` going forward and just use schema examples

openapi/v2/metadata/components/schemas/create_update_custom_object.yaml
```diff
@@ -1,9 +1,9 @@
 type: object
 properties:
-  name:
+  suggested_name:
     type: string
     example: ticket
-    description: In Salesforce, this must end with `__c`.
+    description: The name you'd like to use for the custom object. For some providers, like Salesforce, we might mutate the name, like append `__c`.
```
tomkit, line 6: Can we enumerate the known ones today (Salesforce, Hubspot)

openapi/v2/actions/components/schemas/objects/entity_or_object_record.yaml
albertyfwu: looks like we don't use this anywhere. going to delete this yaml

openapi/v2/actions/components/schemas/objects/entity_or_object.yaml
albertyfwu: this file not used either. deleting.

PR #1376, opened 2023-08-05
fix: FE controlled component for entity mappings, openapi example id, hubspot custom object lookup

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


apps/mgmt-ui/src/pages/applications/[applicationId]/customers/[customerId]/connections/[providerName]/index.tsx
```diff
@@ -357,7 +351,7 @@ function EntityObjectMapping({ entity, customerId, providerName, object, setObje
             options={objectOptions}
             groupBy={(option) => option.type}
             getOptionLabel={(option) => option.name}
-            defaultValue={object}
+            value={object}
```
Comment (albertyfwu) [inline apps/mgmt-ui/src/pages/applications/[applicationId]/customers/[customerId]/connections/[providerName]/index.tsx:354]: using as controlled to avoid issue with `object` changing

packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -587,8 +587,7 @@ class HubSpotClient extends AbstractCrmRemoteClient {
     modifiedAfter?: Date,
     heartbeat?: () => void
   ): Promise<Readable> {
-    // Look up the objectTypeId given the object name
-    const objectTypeId = await this.#getObjectTypeIdByCustomObjectName(object);
+    const objectTypeId = object;
```
albertyfwu, line 590: custom objects are referenced by id now for hubspot, not name, so no lookup necessary

openapi/v2/mgmt/components/schemas/objects/destination.yaml
```diff
@@ -1,15 +1,16 @@
...
-  config:
-    host: production-db-new.cluster-cdhnnutnlctj.us-west-2.rds.amazonaws.com
-    port: 5432
-    database: postgres_prod_db
-    schema: public
-    user: myuser
+# https://github.com/stoplightio/spectral/issues/2081
+# # this example lives here b/c of the oneOf issue
+# example:
+#   id: 2cce5aed-e8fc-41b2-bff6-6ddbc1f99285
```
albertyfwu, line 4: even when this `id` is unique in all openapi specs, for some reason, there is still issue with "maps to more than one schema" with ajv

asdfryan: @tomkit is this okay?

asdfryan (approved):

PR #1384, opened 2023-08-08
fix: redirect to create-organization when none

tomkit:
Some previous problems with the `switch-organization` page:
- I think we were assuming `organizationList` returns the list of orgs associated with a user, but it will actually return all organizations for an environment
- There was a problem with `useEffect`, `router`, and the self-executing function causing an infinite refresh

What I'm doing now instead:
- Auth flow is: SignUp->CreateOrg->Home
   - Instead of: SignUp->SwitchOrg->CreateOrg->Home
- In the `afterAuth` lifecycle event, if there is no user, redirect them to sign-in
   - If there is a user, but no associated org, redirect them to create an organization
   - If there is an organization, it redirects them to the root page
- The switch-organization page isn't used right now, instead when the user is authenticated, they can go to the Profile/Account page to switch organizations

Some other misc fixes:
- Corresponding URL for Clerk production vs development for hosted Account and Organization pages
- Remove `afterSignUpUrl` because I think it gets pre-empted with `afterAuth` in the auth lifecycle

## Test Plan

- Signed up as a new user
- Signed in with an existing user with 1 org
- Signed in with an existing user with 2 orgs
- Signed in with an existing user and no org
- Signed out and in
- Switched organizations for user with 2 orgs

## Deployment instructions


apps/mgmt-ui/src/middleware.ts
```diff
@@ -36,11 +32,12 @@ const cloudMiddleware = authMiddleware({
       return NextResponse.redirect(signInUrl);
     }
 
-    // redirect them to organization selection page
-    if (userId && !orgId && request.nextUrl.pathname !== '/switch-organization') {
-      const orgSwitcher = new URL('/switch-organization', request.url);
-      return NextResponse.redirect(orgSwitcher);
+    // redirect them to create an organization if they don't have one
+    if (userId && !orgId && request.nextUrl.pathname !== '/create-organization') {
```
asdfryan, line 36: I don't this is quite right? `!orgId` means clerk did not pass back an orgId (i.e. they don't have an active org) not that they don't have an org.

apps/mgmt-ui/src/pages/create-organization/[[...index]].tsx
```diff
@@ -8,9 +10,28 @@ export const getServerSideProps: GetServerSideProps = async () => {
...
+  const router = useRouter();
+  const { switcher } = router.query;
   return (
     <div className="m-auto">
-      <CreateOrganization afterCreateOrganizationUrl="/" />
+      {switcher === '1' ? (
+        <div className="flex flex-col items-center gap-6">
+          <Paper className="px-8 py-10 gap-6 flex flex-col" elevation={1}>
+            <Typography variant="h5">Switch organization</Typography>
+            <OrganizationSwitcher afterSwitchOrganizationUrl="/" />
```
Comment (lucasmarshall) [inline apps/mgmt-ui/src/pages/create-organization/[[...index]].tsx:21]:
```suggestion
            <OrganizationSwitcher afterSwitchOrganizationUrl="/" hidePersonal={true} />
```

lucasmarshall (changes requested):

lucasmarshall (approved):

PR #1394, opened 2023-08-09
docs: Magic Link documentation

asdfryan:


albertyfwu (approved):

docs/sidebars.js
```diff
@@ -99,6 +99,11 @@ const sidebars = {
       label: 'Managed authentication',
       type: 'doc',
     },
+    {
+      id: 'platform/magic-link',
```
albertyfwu, line 103: should this be nested in managed auth?

asdfryan: I don't think so -- managed auth is a separate page and not a category page. (Also it makes Magic Links more discoverable in the sidebar)

PR #1395, opened 2023-08-09
chore: update apollo error message to be a little more helpful

asdfryan:


albertyfwu (approved):

packages/core/remotes/impl/apollo/index.ts
```diff
@@ -370,6 +370,9 @@ class ApolloClient extends AbstractEngagementRemoteClient {
         headers: this.#headers,
       }
     );
+    if (response.data.contacts.length === 0) {
```
albertyfwu, line 373: actually, does apollo return any kind of error message in the response? if so, we should use that

asdfryan: No.

In fact they actually return 200 (success) even though they don't actually add the contact to the sequence.

PR #1405, opened 2023-08-10
feat: accept JSON body for passthrough

asdfryan:


packages/core/remotes/impl/apollo/index.ts
```diff
@@ -89,7 +89,12 @@ class ApolloClient extends AbstractEngagementRemoteClient {
       return await super.sendPassthroughRequest({ ...request, query: { ...request.query, api_key: this.#apiKey } });
     }
     if (request.method === 'POST' || request.method === 'PATCH' || request.method === 'PUT') {
-      const bodyJson = request.body ? JSON.parse(request.body as string) : {};
+      let bodyJson: Record<string, unknown> = {};
+      if (typeof request.body === 'string') {
+        bodyJson = JSON.parse(request.body);
```
albertyfwu, line 94: do we actually need to call `JSON.parse()`? does axios not handle it correctly otherwise?

also, i don't think we want to do that, because it's possible that certain APIs aren't expecting JSON payloads. What if the body is literally supposed to be a string?

albertyfwu (commented):
i don't think we should do JSON.parse

albertyfwu (approved):

packages/core/remotes/impl/apollo/index.ts
asdfryan: We do this because we have to inject the api key into the body for Apollo

PR #1415, opened 2023-08-11
feat: add dynamic config

tomkit:
## Test Plan

[Describe test plan here]

## Deployment instructions


apps/mgmt-ui/src/pages/api/internal/env.ts
```diff
@@ -1,15 +1,35 @@
...
+// NOTE: the env variables passed back here are public
+export type SupagluePublicNextEnv = {
+  API_HOST: string;
+  IS_CLOUD: boolean;
+  CLERK_ACCOUNT_URL: string;
+  CLERK_ORGANIZATION_URL: string;
+  homeCtaButtonConfig: HomeCtaButton;
+};
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse<SupagluePublicNextEnv>) {
```
tomkit, line 16: We should make this getServerSideProps

PR #1425, opened 2023-08-14
fix: schemas with special characters in their names, webhook shape

lucasmarshall:
Fixes some issues I noticed while doing the dogfooding


asdfryan (approved):

tomkit:
<3 dogfood

PR #1445, opened 2023-08-16
chore: add additional context to entity syncs

lucasmarshall:


asdfryan (approved):

asdfryan (approved):

packages/sync-workflows/workflows/run_object_sync.ts
```diff
@@ -89,15 +89,19 @@ export async function runObjectSync({ syncId, connectionId, category }: RunObjec
         errorMessage,
       });
     } else {
+      const {
+        entity: { name: entityName },
+      } = await getEntity({ entityId: sync.entityId });
```
albertyfwu, line 94: @lucasmarshall after deploying to prod we should make sure to kill any workflows that get stuck here, since there will be a non-determinism issue in temporal with the new activity call here

packages/sync-workflows/activities/maybe_send_sync_finish_webhook.ts
```diff
@@ -50,17 +51,19 @@ export function createMaybeSendSyncFinishWebhook({
         result: status === 'SYNC_SUCCESS' ? 'SUCCESS' : 'ERROR',
```
albertyfwu, line 51: for the payload above, won't this send the `runId` as `id`? Could that be misinterpreted as `syncId` instead of `syncrunId`?

openapi/v2/mgmt/webhooks/sync.complete.yaml
```diff
@@ -35,9 +40,58 @@ post:
...
+                - object
+            - type: object
+              title: Entity Sync Complete
+              properties:
+                webhook_event_type:
+                  type: string
+                  enum:
+                    - sync.complete
+                  example: sync.complete
+                id:
```
albertyfwu, line 84: should we return `run_id` to make it clear that this is for the run and not the sync?

lucasmarshall: Sure.

PR #1450, opened 2023-08-16
fix: number of employees should not exceed 7B

asdfryan:


packages/core/remotes/impl/hubspot/mappers.ts
```diff
@@ -60,14 +62,19 @@ export const fromHubSpotCompanyToAccount = ({
       ]
     : [];
 
+  let numberOfEmployees = properties.numberofemployees ? parseInt(properties.numberofemployees) : null;
```
albertyfwu, line 65: what if we just reject the request altogether?

albertyfwu (approved):

lucasmarshall (approved):

packages/core/remotes/impl/hubspot/mappers.ts
asdfryan: Well that is what is happening today, and syncs are failing.

PR #1460, opened 2023-08-17
fix: don't send api keys in payloads to sentry

lucasmarshall:


packages/core/lib/logger.ts
```diff
@@ -39,6 +39,13 @@ if (sentryEnabled) {
 
       scope.setUser({ id: data.orgId as string });
     },
+    beforeSend(event) {
+      if (event.request?.headers) {
+        delete event.request.headers['X-Api-Key'];
```
albertyfwu, line 44: Is this case sensitive?

albertyfwu (approved):

packages/core/lib/logger.ts
lucasmarshall: Yes.

PR #1461, opened 2023-08-18
fix: normalize table names

lucasmarshall:
I checked prod and this shouldn't cause any issues with any current tables.

### DIFF
--- a/packages/utils/slugify.ts
+++ b/packages/utils/slugify.ts
@@ -1,0 +1,5 @@
+import slugify from 'slugify';
+
+export function slugifyForTableName(value: string): string {
+  return slugify(value, { replacement: '_', lower: true, strict: true });
+}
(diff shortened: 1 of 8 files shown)


packages/utils/slugify.ts
```diff
@@ -0,0 +1,5 @@
+import slugify from 'slugify';
+
+export function slugifyForTableName(value: string): string {
+  return slugify(value, { replacement: '_', lower: true, strict: true });
```
albertyfwu, line 4: does `replacement` replace spaces with `_`? we currently force entity names to be `A-Za-z0-9_`, so i don't think spaces can be present. Anyways, the lowercasing looks good to me.

albertyfwu (approved):

packages/utils/slugify.ts
lucasmarshall: Yes it does.

PR #1462, opened 2023-08-18
chore: add unit tests for CRM mappers and some util fns

asdfryan:


.husky/pre-commit
```diff
@@ -3,3 +3,4 @@
 
 yarn turbo run lint-staged "--filter=[HEAD^1]"
 yarn turbo run generate "--filter=!salesforce-pub-sub"
+yarn turbo run test "--filter=@supaglue/core"
```
asdfryan, line 6: @lucasmarshall is this right?

packages/core/remotes/impl/ms_dynamics_365_sales/mappers.ts
```diff
@@ -516,11 +516,11 @@ export const fromDynamicsLeadToRemoteLead = (
     dynamicsLead.address2_country
   ) {
     const addressType =
-      dynamicsLead.address1_addresstypecode === 1
+      dynamicsLead.address2_addresstypecode === 1
```
asdfryan, line 519: This is a bug fix

packages/core/remotes/impl/pipedrive/mappers.ts
```diff
@@ -250,11 +256,16 @@ export const toPipedrivePersonCreateParams = (params: ContactCreateParams, field
...
+};
+
+export const toPipedrivePersonUpdateParams = (params: ContactCreateParams, fields: PipedriveObjectField[]) => {
+  const name = getFullName(params.firstName, params.lastName);
   const mappedCustomFields = params.customFields
     ? mapCustomFieldsFromNamesToKeys(params.customFields, fields)
     : undefined;
   return {
-    name,
+    name: name === null ? undefined : name,
```
asdfryan, line 268: Another bugfix

packages/core/remotes/impl/salesforce/mappers.ts
```diff
@@ -258,17 +258,20 @@ export const fromSalesforceLeadToLead = (
...
-        street1: record.Street ?? null,
-        street2: null,
-        city: record.City ?? null,
-        state: record.State ?? null,
-        country: record.Country ?? null,
-        postalCode: record.PostalCode ?? null,
-        addressType: 'primary',
-      },
-    ],
+    addresses:
```
asdfryan, line 261: Another bugfix

.husky/pre-commit
lucasmarshall: Actually, we should probably change some of this.

packages/core/remotes/impl/salesforce/mappers.ts
```diff
@@ -258,17 +258,20 @@ export const fromSalesforceLeadToLead = (
...
+      record.Street || record.City || record.State || record.Country || record.PostalCode
```
albertyfwu, line 262: what if the user wants to explicitly set `PostalCode` to empty string?

asdfryan: This code change makes it so that if all of the fields are empty (null or empty string or otherwise), we don't actually add an Address object. It used to be that we'd add a primary address object no matter, even if all fields were empty.

lucasmarshall (approved):

PR #1464, opened 2023-08-21
fix: don't run sync reconciler in transaction

albertyfwu:
Also adding some logs to see if there is prisma latency/timeout issue.

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/db/prisma/migrations/20230821174715_index/migration.sql
+++ b/packages/db/prisma/migrations/20230821174715_index/migration.sql
@@ -1,0 +1,2 @@
+-- CreateIndex
+CREATE INDEX "connections_provider_id_idx" ON "connections"("provider_id");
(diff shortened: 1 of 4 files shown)


packages/db/prisma/migrations/20230821174715_index/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- CreateIndex
+CREATE INDEX "connections_provider_id_idx" ON "connections"("provider_id");
```
albertyfwu, line 2: just try random things to improve performance

albertyfwu: if we actually want to improve performance, there is a lot more work to do, but let's try the simple things first

lucasmarshall (approved):

PR #1474, opened 2023-08-23
fix: issue with mapping addresses in salesforce mapper

lucasmarshall:


turbo.json
lucasmarshall, line 1: Just some other things I noticed.

packages/core/remotes/impl/salesforce/mappers.ts
```diff
@@ -66,8 +66,8 @@ export const fromSalesforceAccountToAccount = (record: Record<string, any>): Acc
           street2: null,
           city: record.BillingCity ?? null,
           state: record.BillingState ?? null,
-          postalCode: record.postalCode ?? null,
-          country: record.country ?? null,
+          postalCode: record.BillingPostalCode ?? null,
```
albertyfwu, line 69: hm, is `postalCode` and `country` useful for anything?

albertyfwu (approved):

packages/core/remotes/impl/salesforce/mappers.ts
lucasmarshall: They don't exist.

PR #1478, opened 2023-08-25
chore: new supaglue pagination api destination

asdfryan:


openapi/v2/mgmt/components/schemas/objects/destination.yaml
```diff
@@ -11,6 +11,28 @@ example:
...
+    properties:
+      id:
+        type: string
+        example: d455d20b-f6dc-4bc3-ab14-b4f21c4b4835
+      application_id:
+        type: string
+        example: 726fb798-d854-4c59-9a23-57e3e2f73eae
+      name:
+        type: string
+        example: My Postgres Destination
```
lucasmarshall, line 25: Should we allow a name? I think that we probably should set one of our own?

asdfryan: Good point

lucasmarshall (approved):

PR #1490, opened 2023-08-30
docs: salesforce list apis

asdfryan:


openapi/v2/data/openapi.yaml
```diff
@@ -26,8 +26,12 @@ tags:
 paths:
   '/salesforce/contact':
```
albertyfwu, line 27: should this be pluralized too?

albertyfwu (approved):

openapi/v2/data/openapi.yaml
asdfryan: yep youre right

PR #1491, opened 2023-08-30
chore: enrichment openAPI spec and docs

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


openapi/v2/enrichment/components/schemas/objects/enriched_company.yaml
```diff
@@ -0,0 +1,23 @@
+type: object
+properties:
+  name:
```
asdfryan, line 3: Can you come up with realistic examples for these? I feel it makes our docs better.

asdfryan (approved):

openapi/v2/enrichment/components/schemas/objects/enriched_company.yaml
albertyfwu: i did provide example at the top-level (see `paths/persons.yaml`). Isn't that where we are doing examples now? @tomkit

PR #1494, opened 2023-08-30
feat: implement supaglue managed destination

asdfryan:
Tested locally

image


packages/core/destination_writers/supaglue.ts
```diff
@@ -0,0 +1,318 @@
...
+async function createPartitionIfNotExists(client: PoolClient, tableName: string, customerId: string) {
+  const partitionName = `${tableName}_${convertDashesToUnderscores(customerId)}`;
+
+  // Check if the partition already exists
+  const checkQuery = `SELECT to_regclass('${partitionName}')`;
+  const checkResult = await client.query(checkQuery);
+
+  if (checkResult.rows[0].to_regclass === null) {
+    // If partition doesn't exist, create it
+    const createPartitionQuery = `CREATE TABLE ${partitionName} PARTITION OF ${tableName} FOR VALUES IN ('${customerId}')`;
```
albertyfwu, line 38: why don't we create partition on `(application_id, provider_name, customer_id)`?

albertyfwu: what are the pros and cons of partitioning manually by schema and by using native pg feature?

asdfryan: the provider name is already part of the table name (that's what we already do today), and application is schema

albertyfwu: why not just have a single table that stores everything, but each partition is by `(application_id, provider_name, customer_id)`? isn't that easier to manage for schema migrations, etc.? What is the advantage of doing manual partitioning? I actually am not familiar with native partitioning, so I'd like to know the pros/cons.

asdfryan: as far as advantages of partitioning vs namespacing -- not sure. The only thing I can think of is you only have to manage one schema instead of many per customer. Querying across customers is also possible -- though not sure when we'd do that.

asdfryan: well the partitions must all maintain the same schema, so you can't partition by `provider_name` anyway (since salesforce / hubspot may have different schemas eventually, not sure).

As far as application_id -- not sure but your point about schema changes is a good one. I'm not sure whether or not we would ever do different schemas per application, but that could be a good reason.

asdfryan: @george-xing ? Would we ever deviate from our current schema (where we just store `_supaglue_raw_data` as JSON today)? Would we ever have different schemas per provider?

asdfryan: Evidently you cannot partition by multiple columns. So this may be moot?

packages/core/destination_writers/supaglue.ts
```diff
@@ -0,0 +1,318 @@
...
+import { getPgPool, logger } from '../lib';
+import type { WriteCommonObjectRecordsResult, WriteEntityRecordsResult, WriteObjectRecordsResult } from './base';
+import { BaseDestinationWriter } from './base';
+
+function convertDashesToUnderscores(input: string): string {
+  return input.replace(/-/g, '_');
+}
+
+async function createPartitionIfNotExists(client: PoolClient, tableName: string, customerId: string) {
+  const partitionName = `${tableName}_${convertDashesToUnderscores(customerId)}`;
```
albertyfwu, line 30: Is this sufficient? What other characters are disallowed by postgres? Shouldn't we reduce to that set?

packages/core/destination_writers/supaglue.ts
```diff
@@ -0,0 +1,318 @@
...
+  async #writeRecords(
+    { providerName, customerId, applicationId }: ConnectionSafeAny,
+    table: string,
+    inputStream: Readable,
+    heartbeat: () => void,
+    childLogger: pino.Logger
+  ): Promise<WriteObjectRecordsResult> {
+    const schema = convertDashesToUnderscores(applicationId);
+    const qualifiedTable = `${schema}.${table}`;
+    const tempTable = `"temp_${table}"`;
```
albertyfwu, line 122: Shouldn't we not use quotes here?

packages/core/destination_writers/supaglue.ts
```diff
@@ -0,0 +1,318 @@
...
+  "_supaglue_id" TEXT NOT NULL,
+  "_supaglue_emitted_at" TIMESTAMP(3) NOT NULL,
+  "_supaglue_last_modified_at" TIMESTAMP(3) NOT NULL,
+  "_supaglue_is_deleted" BOOLEAN NOT NULL,
+  "_supaglue_raw_data" JSONB NOT NULL
+
+  ${
+    temp
+      ? ''
+      : ', PRIMARY KEY ("_supaglue_application_id", "_supaglue_provider_name", "_supaglue_customer_id", "_supaglue_id")'
```
albertyfwu, line 301: one could argue that we don't need to have application_id, provider_name here because we are putting it in the schema / table names, but this is fine too.

packages/core/destination_writers/supaglue.ts
```diff
@@ -0,0 +1,318 @@
...
+  }
+    
+)${temp ? ';' : ' PARTITION BY LIST ( _supaglue_customer_id );'}`;
```
albertyfwu, line 304: is partition efficient if the partition key is not what the primary key is?

asdfryan: Yes, as long as the partition key is a subset of the primary key.

packages/core/destination_writers/supaglue.ts
asdfryan: I copied this from already existing code.

packages/core/destination_writers/supaglue.ts
albertyfwu: can it be any subset, or does it have to be the first column in the primary key? e.g. for indices, you can't use it if it's not the first column.

asdfryan: I can't find documentation that you _can't_ do it, but I also tested it empirically, so I assume it's fine.

packages/core/destination_writers/supaglue.ts
asdfryan: Sure.

albertyfwu (approved):

PR #1505, opened 2023-08-31
chore: code to reconcile ConnectionSyncConfigChange

albertyfwu:
This works for standard objects on Connection.ConnectionSyncConfig, but does not implement custom object syncs yet.

## Test Plan

[Describe test plan here]

## Deployment instructions


packages/sync-workflows/services/sync_service.ts
```diff
@@ -122,8 +335,16 @@ export class SyncService {
       const relevantConnections = await this.#connectionService.getSafeByProviderId(syncConfig.providerId);
 
       const syncArgs: Prisma.SyncCreateManyInput[] = [];
+      const connectionIdsToUpdate: string[] = [];
 
       for (const connection of relevantConnections) {
+        // If connection already has its own ConnectionSyncConfig, then skip it
```
albertyfwu, line 341: does this look right?

lucasmarshall (approved):

PR #1510, opened 2023-09-01
chore: add error specific to cache invalidation

tomkit:
## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/errors.ts
+++ b/packages/core/errors.ts
@@ -11,6 +11,12 @@ export class NotImplementedError extends SGError {
   }
 }
 
+export class CacheInvalidationError extends SGError {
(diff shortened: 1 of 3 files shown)


packages/core/errors.ts
```diff
@@ -11,6 +11,12 @@ export class NotImplementedError extends SGError {
   }
 }
 
+export class CacheInvalidationError extends SGError {
```
lucasmarshall, line 14: This will still be a 500. Is that what we want?

albertyfwu (approved):
this works, but couldn't we also just `try/catch` in a single place in `BaseDestinationWriter` instead?

PR #1512, opened 2023-09-01
chore: make hubspot association keys singular instead of plural

albertyfwu:
Instead of 

```
{
  ...
  "properties": {
    ...
  },
  "associations": {
    "contacts": [
      "101"
    ],
    "2-17410842": [
      "7999197075",
      "7999197075"
    ]
 }
}
```

do this

```
{
  ...
  "properties": {
    ...
  },
  "associations": {
    "contact": [
      "101"
    ],
    "2-17410842": [
      "7999197075",
      "7999197075"
    ]
 }
}
```

## Test Plan

[Describe test plan here]

## Deployment instructions

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/core/remotes/impl/hubspot/index.ts
+++ b/packages/core/remotes/impl/hubspot/index.ts
@@ -157,6 +157,23 @@ const hubspotStandardObjectTypeToPlural: Record<HubSpotStandardObjectType, strin
   task: 'tasks',
 };
 
+const hubspotStandardObjectPluralizedToType: Record<string, HubSpotStandardObjectType> = {
+  companies: 'company',
+  contacts: 'contact',
+  deals: 'deal',
+  'line items': 'line_item',
+  products: 'product',
+  tickets: 'ticket',
+  quotes: 'quote',
+  calls: 'call',
+  communications: 'communication',
+  emails: 'email',
+  meetings: 'meeting',
+  notes: 'note',
+  'postal mails': 'postal_mail',
+  tasks: 'task',
+};
+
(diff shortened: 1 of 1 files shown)


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -753,20 +770,24 @@ class HubSpotClient extends AbstractCrmRemoteClient {
         ...response.data,
         results: response.data.results.map(({ associations, ...rest }) => ({
           ...rest,
-          associations: Object.entries(associations ?? {}).reduce((acc, [associatedObjectType, { results }]) => {
-            // If associatedObjectType is for a standard object, it will be pluralized
-            if (HUBSPOT_STANDARD_OBJECT_TYPES_PLURALIZED.includes(associatedObjectType)) {
-              acc[associatedObjectType] = results.map(({ id }) => id);
+          associations: Object.entries(associations ?? {}).reduce((acc, [associatedObjectTypeKey, { results }]) => {
```
asdfryan, line 773: Are we going to do this for all the other object types too? (i.e. common objects, entities, custom objects etc)

asdfryan: We should keep them consistent IMO

albertyfwu: we don't sync associations for common objects. also, these methods are the same ones used for entities, custom object, and standard objects

asdfryan (approved):

PR #1521, opened 2023-09-05
feat: upsert for contacts and accounts

asdfryan:
Tested locally

image
Screenshot 2023-09-05 at 2 49 03 PM


openapi/v2/crm/openapi.yaml
```diff
@@ -28,10 +28,14 @@ servers:
 paths:
   '/accounts':
     $ref: paths/accounts.yaml
+  '/accounts/upsert':
```
albertyfwu, line 31: i thought we want to follow convention that non-REST-ful things are prefixed with `_`

albertyfwu (approved):

openapi/v2/crm/openapi.yaml
asdfryan: sure

PR #1539, opened 2023-09-07
feat: hubspot form apis

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -415,6 +445,81 @@ class HubSpotClient extends AbstractCrmRemoteClient {
...
+    return response.data.map((form) => ({
+      id: form.guid,
+      name: form.name,
+      createdAt: new Date(form.createdAt),
+      updatedAt: new Date(form.updatedAt),
+      rawData: form,
+    }));
+  }
+
+  public async getFormFields(formId: string): Promise<FormField[]> {
```
tomkit, line 502: can you also name these to be associated with Marketing Automation (like we do below with remote client?)

tomkit (approved):

PR #1552, opened 2023-09-09
feat: Sequence and sequence create endpoints

asdfryan:
Tested locally

Screenshot 2023-09-08 at 5 06 19 PM
Screenshot 2023-09-08 at 5 05 15 PM


openapi/v2/data/components/schemas/objects/salesforce/account.yaml
```diff
@@ -62,6 +62,7 @@ properties:
   LastActivityDate:
     type: string
     nullable: true
+    example: '2022-02-27T00:00:00Z'
```
asdfryan, line 65: This is so we get stable docs

albertyfwu (approved):

PR #1570, opened 2023-09-12
feat: return object properties for intercom contacts & companies

albertyfwu:
also remove support for common object properties

## Test Plan

[Describe test plan here]

## Deployment instructions


packages/types/property.ts
```diff
@@ -1,4 +1,8 @@
 export type Property = {
   id: string;
   label: string;
+
+  // this is passed back directly from the provider
+  // we do not have our own enum mapping here
+  type?: string;
```
asdfryan, line 7: Where are we providing the types for common objects?

albertyfwu: we're only providing the types that our customers need at the moment. for common objects that have been implemented, we're not providing the types. we'll do that later, if necessary.

asdfryan (approved):

PR #1576, opened 2023-09-13
chore: clarify that bg dataset and pg schema can be overriden

albertyfwu:
## Test Plan

[Describe test plan here]

## Deployment instructions


apps/mgmt-ui/src/components/connectors/destination/BigQueryDestinationDetailsPanel.tsx
```diff
@@ -225,7 +225,7 @@ export default function BigQDestinationDetailsPanel({ isLoading }: BigQueryDesti
             size="small"
             label="Dataset"
             variant="outlined"
-            helperText="This is the dataset where tables will be written into (it must already exist)."
+            helperText="This is the dataset where tables will be written into (it must already exist). You can override this dataset per connection via the API later."
```
tomkit, line 228: `You can override this dataset per connection via the API later.` --> `Use the Management ConnectionSyncConfigs API to override this.`

tomkit (approved):

PR #1579, opened 2023-09-13
fix: only cache invalidate if sync exists

asdfryan:


packages/core/services/common_objects/engagement/common_object_service.ts
```diff
@@ -65,6 +73,11 @@ export class EngagementCommonObjectService {
     const res = await remoteClient.updateCommonObjectRecord(type, params);
     end();
 
+    const shouldCacheInvalidate = await this.#shouldPerformCacheInvalidation(connection, type);
```
albertyfwu, line 76: optional: seems like we can abstract out the entire cache invalidation code, and not just `shouldPerformCacheInvalidation`

albertyfwu (approved):

PR #1584, opened 2023-09-14
fix: two run object sync changes

tomkit:
- bump activity timeout from 10s to 12s
- try to log the inner most cause first

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/packages/sync-workflows/workflows/run_object_sync.ts
+++ b/packages/sync-workflows/workflows/run_object_sync.ts
@@ -283,9 +283,11 @@ async function doFullThenIncrementalSync(sync: FullThenIncrementalSync): Promise
 
 const getErrorMessageStack = (err: Error): { message: string; stack: string } => {
   if (err instanceof ActivityFailure) {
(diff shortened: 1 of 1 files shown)


packages/sync-workflows/workflows/run_object_sync.ts
```diff
@@ -18,7 +18,7 @@ const { syncObjectRecords, syncEntityRecords } = proxyActivities<ReturnType<type
 const { getSync, updateSyncState, clearSyncArgsForNextRun, logSyncStart, logSyncFinish, getEntity } = proxyActivities<
   ReturnType<typeof createActivities>
 >({
-  startToCloseTimeout: '10 second',
+  startToCloseTimeout: '12 second',
```
albertyfwu, line 21: Why does 10->12 seconds fix issues? What's the root cause of timeouts we are encountering?

albertyfwu (approved):

packages/sync-workflows/workflows/run_object_sync.ts
tomkit: Removed

PR #1587, opened 2023-09-14
docs: more soft delete and incremental clarity. apollo limitations

tomkit:
## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/docs/docs/providers/apollo.md
+++ b/docs/docs/providers/apollo.md
@@ -13,23 +13,25 @@ description: ''
 
 Apollo is both a sales engagement connector and a data enrichment provider.
 
-| Feature                              | Available   |
-| ------------------------------------ | ----------- |
-| Authentication (`api_key`)           | Yes         |
-| Managed syncs                        | Yes         |
-| &nbsp;&nbsp;&nbsp; Sync strategies   | `full only` |
-| Unified API                          | Yes         |
-| &nbsp;&nbsp;&nbsp; Data invalidation | No          |
-| Real-time events                     | No          |
-| Passthrough API                      | Yes         |
+| Feature                              | Available      |
+| ------------------------------------ | -------------- |
+| Authentication (`api_key`)           | Yes            |
+| Managed syncs                        | Yes            |
+| &nbsp;&nbsp;&nbsp; Sync strategies   | (listed below) |
+| Unified API                          | Yes            |
+| &nbsp;&nbsp;&nbsp; Data invalidation | Yes            |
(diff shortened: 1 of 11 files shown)


docs/docs/providers/apollo.md
```diff
@@ -39,6 +41,16 @@ N/A
...
 N/A
 
+## Other limitations
+
+Date: 9/14/2023
+
+Apollo has a few hard limitations with its API:
+
+- You cannot sync more than 10,000 records per Object type.
+- You cannot sync more than 10,000 records per hour.
```
albertyfwu, line 51: do we want to mention that this depends on free/paid/custom?

docs/docs/providers/hubspot.md
```diff
@@ -13,43 +13,58 @@ description: ''
...
+| Object  | Soft delete supported | Sync strategy           |
+| ------- | --------------------- | ----------------------- |
+| Company | Yes                   | `full then incremental` |
+| Contact | Yes                   | `full then incremental` |
+| Deal    | Yes                   | `full then incremental` |
+| User    | Yes                   | `full only`             |
 
 :::info
-Note: if only associations change in between syncs, Supaglue syncs will not reflect it until the relevant common object is modified.
+Note: Supaglue does not sync association changes until the relevant object gets modified in the remote provider.
```
albertyfwu, line 36: should this be `until a relevant object`, since there are 2 candidate objects?

albertyfwu (approved):

docs/docs/providers/hubspot.md
```diff
@@ -13,43 +13,58 @@ description: ''
...
-- Company
-- Contact
-- Deal
-- User ([*] Sync strategy: `full only`)
+| Object  | Soft delete supported | Sync strategy           |
+| ------- | --------------------- | ----------------------- |
+| Company | Yes                   | `full then incremental` |
+| Contact | Yes                   | `full then incremental` |
+| Deal    | Yes                   | `full then incremental` |
+| User    | Yes                   | `full only`             |
```
albertyfwu, line 33: hubspot api only allows listing all users, so we list all users. however, we do a client-side filtering so that only newly-updated users are emitted to the destination.

We did this for two reasons:
1. Reduce load on destination
2. Respect invariants and function names where consumers expect the stream to only contain new records.

We probably should adopt a common convention for describing this client-side filtering for other providers too.

Also, for most providers, we are just lazy and do a full refresh even if we can do client-side filtering. We should address that down the road.

PR #1593, opened 2023-09-14
fix: delete temp table after standard obj syncs

asdfryan:
We do it for common obj but not for standard / entities


albertyfwu (approved):

tomkit (approved):
Ty, we might want to also delete at the start of a sync to address dangling temp tables

PR #1594, opened 2023-09-14
chore!: don't write mapped_data for users unless whitelisted

tomkit:
## Test Plan

[Describe test plan here]

## Deployment instructions


asdfryan (commented):
I thought we were continuing to support schemas and entities?

albertyfwu (approved):

PR #1595, opened 2023-09-15
feat: delete records in full syncs

asdfryan:
We can do this by diffing temp table and real table for full syncs only.

This PR also removes the S3 destination


tomkit:
Ty - 

Do we want to not do this for Apollo? It might cause arbitrary records to be soft deleted if users are hitting the 10K record limit?

asdfryan:
> Ty -
> 
> Do we want to not do this for Apollo? It might cause arbitrary records to be soft deleted if users are hitting the 10K record limit?

Good point. I can skip this step for apollo.

packages/core/destination_writers/util.ts
```diff
@@ -46,3 +46,7 @@ const snakecasedKeysMapperByCommonObjectType: {
     user: toSnakecasedKeysEngagementUser,
   },
 };
+
+export const shouldDeleteRecords = (isFullSync: boolean, providerName: ProviderName): boolean => {
+  return isFullSync && providerName !== 'apollo';
```
albertyfwu, line 51: should we document this?

albertyfwu: also we should introduce some concept to represent this in the future to avoid introducing even more leaky abstractions.

imo destinations should never know about providers

albertyfwu (approved):

packages/core/destination_writers/util.ts
asdfryan: Sure -- I changed the interface to pass in `diffAndDeleteRecords`

PR #1608, opened 2023-09-18
chore: update turbo and prisma

lucasmarshall:


packages/db/package.json
```diff
@@ -6,12 +6,12 @@
   "index": "dist/index.js",
   "types": "index.ts",
   "dependencies": {
-    "@prisma/client": "^5.1.1",
+    "@prisma/client": "^5.3.1",
```
lucasmarshall, line 9: There's a segfault issue on arm64-based linux distros that is fixed by this version.

asdfryan (approved):

PR #1615, opened 2023-09-19
feat: sync common objects to managed destination

asdfryan:


tomkit (approved):
approved assuming we resolve dependency/import issues

nit: consider moving the postgres-related utils into a `postgres_util` since they're only relevant for postgres (or SQL) destinations

PR #1616, opened 2023-09-19
feat: Unified CRM list API

asdfryan:
Also includes some QoL improvements to pagination API:
- Return more useful error message when syncs are not properly set up
- Make default page size 10 instead of 1000

Screenshot 2023-09-19 at 2 47 32 PM
Screenshot 2023-09-19 at 2 40 53 PM


george-xing:
In the open api docs, "List Contacts" is incorrectly labeled as "List Accounts"

tomkit (approved):

PR #1618, opened 2023-09-20
feat!: (step 1/2) add cuid to object_sync_runs, double-write runid and cuid, cuid…

tomkit:
… backfill script

Overview:
- Move to cuid from uuid for `object_sync_runs.id` to utilize Prisma cursor pagination. We need the sort key and pagination key to be the same to avoid a table scan

Step 1:
- This creates the column, double writes to it, and contains a backfill script for existing sync runs
   - We will only look back 3 days for the backfill
   - We will lose the original start_timestamp order because it's too expensive to backfill in that order
   - New sync runs will be written in start_timestamp order

Step 2:
- Finish migrating `object_sync_runs.id` to use cuid instead of uuid and utilize it for the sync run mgmt-ui queries

NOTE:
- The mgmt api endpoint hasn't been updated

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/api/routes/internal/_multitenant_backfill/index.ts
+++ b/apps/api/routes/internal/_multitenant_backfill/index.ts
@@ -1,0 +1,67 @@
+import { getDependencyContainer } from '@/dependency_container';
+import { logger } from '@supaglue/core/lib';
+import cuid from 'cuid';
+import type { Request, Response } from 'express';
+import { Router } from 'express';
+
+const { prisma } = getDependencyContainer();
+
+export default function init(app: Router): void {
+  const systemRouter = Router();
+
+  systemRouter.post('/_object_sync_runs_cuid', async (req: Request, res: Response) => {
+    let cursor: { id: string } | null = { id: '' };
+
+    const date = new Date();
+    date.setDate(date.getDate() - 7);
+
+    while (cursor !== null) {
+      await prisma.$transaction(
+        async (tx) => {
+          const records = await tx.syncRun.findMany({
+            cursor: cursor && cursor.id ? cursor : undefined,
+            skip: cursor ? 1 : undefined,
+            take: 5000,
+            where: {
+              startTimestamp: {
+                gte: date,
+              },
+            },
+            orderBy: {
+              id: 'asc',
+            },
+          });
+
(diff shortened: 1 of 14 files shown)


apps/api/routes/internal/_multitenant_backfill/index.ts
```diff
@@ -0,0 +1,67 @@
...
+            skip: cursor ? 1 : undefined,
+            take: 5000,
+            where: {
+              startTimestamp: {
+                gte: date,
+              },
+            },
+            orderBy: {
+              id: 'asc',
+            },
```
lucasmarshall, line 32: Why backfill at all if we aren't actually going to be backfilling the correct order?

apps/sync-worker/index.ts
```diff
@@ -106,6 +106,12 @@ async function run() {
       activityInbound: [(ctx) => new ActivityLogInterceptor(ctx)],
       workflowModules: [`${__dirname}/interceptors/workflow_log_interceptor`],
     }),
+    bundlerOptions: {
+      webpackConfigHook(config) {
+        config.target = 'node';
+        return config;
+      },
+    },
```
lucasmarshall, line 114: We should keep an eye out if this causes any issues. I'm not expecting any, but it _is_ a fundamental change to the webpack config.

packages/sync-workflows/activities/get_sync.ts
```diff
@@ -1,4 +1,5 @@
 import type { Sync } from '@supaglue/types/sync';
+import { default as cuid } from 'cuid';
```
lucasmarshall, line 2: Can't this just be:
```suggestion
import cuid from 'cuid';
```

apps/api/routes/internal/_multitenant_backfill/index.ts
tomkit: Good thought. I also thought about it: I think it's worth at least being able to look at logs from the past 3 days (even in the wrong order) vs truncating it all

lucasmarshall: I don't know that any logs with an incorrect order are useful. Logs by definition are time-ordered and they are used to see performance over time.

asdfryan (approved):

PR #1619, opened 2023-09-20
feat!: (step 2/2) rename object_sync_runs cuid to id

tomkit:
fix!: change sync run id pk, drop id col
fix!: change sync run id pk

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/api/routes/internal/_multitenant_backfill/index.ts
+++ b/apps/api/routes/internal/_multitenant_backfill/index.ts
@@ -1,69 +1,6 @@
-import { getDependencyContainer } from '@/dependency_container';
-import { logger } from '@supaglue/core/lib';
-import cuid from 'cuid';
-import type { Request, Response } from 'express';
 import { Router } from 'express';
-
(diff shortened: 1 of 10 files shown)


apps/api/routes/internal/_multitenant_backfill/index.ts
```diff
@@ -1,67 +1,6 @@
...
-import cuid from 'cuid';
-import type { Request, Response } from 'express';
 import { Router } from 'express';
 
-const { prisma } = getDependencyContainer();
-
 export default function init(app: Router): void {
   const systemRouter = Router();
-
-  systemRouter.post('/_object_sync_runs_cuid', async (req: Request, res: Response) => {
```
asdfryan, line 12: Do we no longer need this?

lucasmarshall: We will have already run it by the time we deploy this

lucasmarshall (approved):

PR #1631, opened 2023-09-22
chore: update prepare_release script to support two-step releases

tomkit:
## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/scripts/prepare_release.sh
+++ b/scripts/prepare_release.sh
@@ -2,13 +2,16 @@
 
 set -euo pipefail
 
-if [ -z "${1-}" ]; then
-  echo "usage: $0 <version>"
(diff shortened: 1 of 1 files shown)


scripts/prepare_release.sh
```diff
@@ -18,9 +21,10 @@ fi
 
 echo "Release version: ${VERSION}"
 
-echo "Checking out latest main branch..."
+echo "Resetting to out latest main branch..."
```
lucasmarshall, line 24:
```suggestion
echo "Resetting to latest main branch..."
```

lucasmarshall (changes requested):

PR #1635, opened 2023-09-22
chore: add tsc:watch target

lucasmarshall:
Can run like this to watch a target and its dependencies:

```
yarn tsc:watch --filter="api..."
```


asdfryan (approved):

tonyxiao:
Does turbo allow you do to things like passing `--watch` flag down to each package? this way we can just have a single `yarn tsc` target and add the `--watch` flag as needed?

PR #1640, opened 2023-09-25
chore: don't allow releases if staging checks are failing

lucasmarshall:


asdfryan (commented):
I wonder if this should be run in the `prepare_release.sh` instead of `release.sh`?

lucasmarshall:
> I wonder if this should be run in the `prepare_release.sh` instead of `release.sh`?

It's possible that something in `prepare_release.sh` could cause failures. It's best to check as close to release as we can.

asdfryan (approved):

PR #1641, opened 2023-09-26
fix: remove usage of metadata api for getting and listing custom salesforce objects

lucasmarshall:
## Test Plan

Tested locally with dev salesforce instance

### DIFF
--- a/packages/core/remotes/impl/salesforce/index.ts
+++ b/packages/core/remotes/impl/salesforce/index.ts
@@ -314,10 +314,10 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
 
   public override async listCustomObjects(): Promise<SimpleCustomObject[]> {
     const metadata = await this.#client.describeGlobal();
-    // for some reason, this returns standard objects and external objects too,
+    // this returns standard objects and external objects too,
     // so we need to filter them out
     return metadata.sobjects
-      .filter(({ name }) => name.endsWith('__c'))
+      .filter(({ custom }) => custom)
       .map(({ name }) => ({
         id: name,
         name: name,
--- a/packages/core/remotes/impl/salesforce/index.ts
+++ b/packages/core/remotes/impl/salesforce/index.ts
@@ -313,14 +313,14 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
   }
 
   public override async listCustomObjects(): Promise<SimpleCustomObject[]> {
-    const metadata = await this.#client.metadata.list({ type: 'CustomObject' });
+    const metadata = await this.#client.describeGlobal();
     // for some reason, this returns standard objects and external objects too,
     // so we need to filter them out
-    return metadata
-      .filter(({ fullName }) => fullName.endsWith('__c'))
-      .map(({ fullName }) => ({
-        id: fullName,
-        name: fullName,
+    return metadata.sobjects
+      .filter(({ name }) => name.endsWith('__c'))
+      .map(({ name }) => ({
+        id: name,
+        name: name,
       }));
   }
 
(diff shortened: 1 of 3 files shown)


packages/core/remotes/impl/salesforce/mappers.ts
```diff
@@ -446,77 +446,38 @@ const toSalesforceEmailCreateParams = (emailAddresses?: EmailAddress[]): Record<
...
-        };
-      }),
+      ...salesforceCustomObject.fields
+        .filter((field) => field.name !== nameField.name)
+        .filter((field) => field.type === 'string' || field.type === 'number')
+        .map((field) => {
+          return {
+            keyName: field.name,
+            displayName: field.label,
+            fieldType: field.type as CustomObjectFieldType,
```
tomkit, line 477: I know this is legacy, but are "text" and "number" enough for the customer? cc @george-xing

packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -313,14 +313,14 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
   }
 
   public override async listCustomObjects(): Promise<SimpleCustomObject[]> {
-    const metadata = await this.#client.metadata.list({ type: 'CustomObject' });
+    const metadata = await this.#client.describeGlobal();
```
tomkit, line 316: `const describedObject` or something like that since this is not longer the Metadata API?

packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -313,14 +313,14 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
...
     // for some reason, this returns standard objects and external objects too,
     // so we need to filter them out
-    return metadata
-      .filter(({ fullName }) => fullName.endsWith('__c'))
-      .map(({ fullName }) => ({
-        id: fullName,
-        name: fullName,
+    return metadata.sobjects
+      .filter(({ name }) => name.endsWith('__c'))
```
tomkit, line 320: It looks like `.../sobject/:objectname/describe` `fields` array in the response has `"custom": true | false` -- do we expect all custom fields to end with `__c`?

packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -561,10 +561,10 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
       // Salesforce doesn't actually enforce this, and will just append __c.
       // However, we want to enforce this to avoid confusion when using
       // the custom object name in other places.
-      throw new Error('Custom object id must end with __c');
+      throw new BadRequestError('Custom object id must end with __c');
     }
 
-    const metadata = await this.#client.metadata.read('CustomObject', id);
+    const metadata = await this.#client.describe(id);
```
tomkit, line 567: `const describedObject` or something similar since no longer from Metadata API?

packages/core/remotes/impl/salesforce/mappers.ts
```diff
@@ -446,77 +446,38 @@ const toSalesforceEmailCreateParams = (emailAddresses?: EmailAddress[]): Record<
...
 };
 
-export const toCustomObject = (salesforceCustomObject: SalesforceCustomObject): CustomObject => {
-  if (!salesforceCustomObject.fullName) {
-    throw new Error(`unexpectedly, custom object missing fullName; custom object may not exist`);
-  }
-
-  if (!salesforceCustomObject.nameField || !salesforceCustomObject.nameField.label) {
+export const toCustomObject = (salesforceCustomObject: DescribeSObjectResult): CustomObject => {
+  const nameField = salesforceCustomObject.fields.find((field) => field.nameField);
```
tomkit, line 450: `const field` instead? `nameField` is the name of the field you're filtering on and is a boolean

lucasmarshall: This is to identify the field that is the name field

packages/core/remotes/impl/salesforce/mappers.ts
lucasmarshall: I don't think it's appropriate to change the interface in this PR.

packages/core/remotes/impl/salesforce/index.ts
lucasmarshall: It's still metadata

packages/core/remotes/impl/salesforce/index.ts
lucasmarshall: I can switch to use that, but by convention they must end with that

packages/core/remotes/impl/salesforce/index.ts
lucasmarshall: It's still metadata, so this name applies

tomkit (approved):

PR #1647, opened 2023-09-26
fix: return cache invalidation errors on remote read or db write errors

lucasmarshall:


asdfryan (changes requested):
we also need to update `engagement/common_object_service.ts`

packages/core/services/common_objects/crm/common_object_service.ts
```diff
@@ -95,11 +95,15 @@ export class CrmCommonObjectService {
...
-      await writer.upsertCommonObjectRecord<'crm', T>(connection, objectName, record);
-      end();
+      try {
+        const record = await this.get(objectName, connection, id);
+
+        const end = remoteDuration.startTimer({ operation: 'create', remote_name: destinationType! });
+        await writer.upsertCommonObjectRecord<'crm', T>(connection, objectName, record);
+        end();
+      } catch (err: any) {
+        throw new CacheInvalidationError(err.message, err);
```
tomkit, line 105: I think we should throw our own error message, i.e. what we had before so it's clear it's cache invalidation, but log and pass along the remote provider error into the jsonapi "details" field (which we're doing here in the second argument) for developers

packages/core/destination_writers/mongodb.ts
```diff
@@ -269,7 +268,7 @@ export class MongoDBDestinationWriter extends BaseDestinationWriter {
       );
     } catch (err) {
       childLogger.error({ err }, 'Error upserting common object record');
```
tomkit, line 270: In retrospect: we should log an error here so it's clear that the upsert is causing a cache invalidation issue

tomkit (commented):
Can we do extra testing for this (positive and negative case) b/c it's in the critical path customers are sensitive to changes here

packages/core/services/common_objects/crm/common_object_service.ts
lucasmarshall: Isn't it clear it's a cache invalidation error since the error code is `CACHE_INVALIDATION_ERROR`?

packages/core/destination_writers/mongodb.ts
lucasmarshall: We are already logging that here.

lucasmarshall:
> Can we do extra testing for this (positive and negative case) b/c it's in the critical path and customers are sensitive to changes here

I already did testing here for both cases. What more testing would you like?

packages/core/destination_writers/mongodb.ts
tomkit: I guess the litmus test is: Will we be able to quickly scan Datadog for and differentiate between cache invalidation vs other upsert() errors?

packages/core/services/common_objects/crm/common_object_service.ts
tomkit: Yes, true. That sgtm

tomkit:
> > Can we do extra testing for this (positive and negative case) b/c it's in the critical path and customers are sensitive to changes here
> 
> I already did testing here for both cases. What more testing would you like?

Nice - ty. I'm addin extra diligence here b/c there's a large matrix, it's in the critical path, and customers are sensitive. My only check is: can we do both Engagement and CRM tests (if not already)?

packages/core/destination_writers/mongodb.ts
lucasmarshall: Not unless we pass a flag here that says this upsert is for cache invaldation. Is that what we want?

packages/core/services/common_objects/engagement/common_object_service.ts
```diff
@@ -81,13 +82,17 @@ export class EngagementCommonObjectService {
...
     const [writer, destinationType] = await this.#destinationService.getWriterByProviderId(connection.providerId);
     if (writer) {
-      // TODO: we should move this logic into each individual provider instead of checking apollo here
-      const record =
-        connection.providerName === 'apollo' ? res.record : await remoteClient.getCommonObjectRecord(type, res.id);
-      if (record) {
-        const end = remoteDuration.startTimer({ operation: 'update', remote_name: destinationType! });
-        await writer.upsertCommonObjectRecord<'engagement', T>(connection, type, record);
-        end();
+      try {
```
asdfryan, line 85: Unfortunately you also need to update `create()` as well.

It'd be great to refactor this but it's up to you.

asdfryan (changes requested):

asdfryan (approved):

PR #1648, opened 2023-09-26
chore: default new apps to development environment

lucasmarshall:


asdfryan (approved):

asdfryan (commented):
Does this not need a migration?

lucasmarshall:
> Does this not need a migration?

Good catch. For some reason, I thought defaults were handled in the generated prisma client.

PR #1650, opened 2023-09-26
fix: don't retry on hourly/daily apollo rate limits, or non-rate-limit errors

lucasmarshall:


asdfryan (approved):

tomkit:
Retroactive: rather than this inversion of concerns inside a low level retry lib (that is used in numerous places), can we just duplicate this code and have apollo use its own version? (I think it's the same amount of work, but we don't need to think about apollo in this low level util)

PR #1658, opened 2023-09-27
fix: map appropriate salesforce errors to 400s

lucasmarshall:


tomkit:
Can we be conservative and keep `MALFORMED_ID`, `INVALID_CROSS_REFERENCE_KEY`, DUPLICATE_VALUE` as 500s until we get more data points on them? Otherwise, lgtm

tomkit (approved):

PR #1663, opened 2023-09-28
fix: camelcasing for upsert

asdfryan:
We need to fix this because there's 1 more additional layer of nesting.

### DIFF
--- a/apps/api/routes/crm/v2/account.ts
+++ b/apps/api/routes/crm/v2/account.ts
@@ -20,7 +20,7 @@ import type {
   UpsertAccountRequest,
   UpsertAccountResponse,
 } from '@supaglue/schemas/v2/crm';
-import { camelcaseKeysSansCustomFields } from '@supaglue/utils/camelcase';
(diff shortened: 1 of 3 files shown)


lucasmarshall (approved):

apps/api/routes/crm/v2/account.ts
```diff
@@ -95,11 +95,10 @@ export default function init(app: Router): void {
...
       res: Response<UpsertAccountResponse>
     ) => {
-      const id = await crmCommonObjectService.upsert(
-        'account',
-        req.customerConnection,
-        camelcaseKeysSansCustomFields(req.body)
-      );
+      const id = await crmCommonObjectService.upsert('account', req.customerConnection, {
+        record: camelcaseKeysSansCustomFields(req.body.record),
+        upsertOn: camelcaseKeys(req.body.upsert_on),
```
tomkit, line 100: it might be worth leaving a comment about this

tomkit (approved):

PR #1667, opened 2023-09-28
fix!: make lists endpoint restful. use hubspot v3 contact for list memberhip

tomkit:
- call hubspot v3 contacts batch read after getting the list of contacts from hubspot v1 endpoint so we can use the existing mapper

salesforce: https://gist.github.com/tomkit/696c36f53202332bf814e6a42c336e3d
hubspot: https://gist.github.com/tomkit/ce441a99ccd0be0dc89a728ba2c7e688

## Test Plan

[Describe test plan here]

## Deployment instructions


lucasmarshall (approved):

packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -95,9 +94,9 @@ import { toMappedProperties } from '../../utils/properties';
 import {
   fromHubSpotCompanyToAccount,
   fromHubSpotContactToContact,
+  fromHubSpotContactToContact_v2,
```
tomkit, line 97: Using an underscore here to disambiguate from Hubspot API versions

packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -2103,15 +2102,10 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
           Authorization: `Bearer ${this.#config.accessToken}`,
         },
         params: {
-          property: propertiesToFetch,
+          property: 'hs_object_id',
```
tomkit, line 2105: Just fetch the record ID

PR #1673, opened 2023-09-29
fix: send 400 errors on INVALID_EMAIL errors from hubspot

lucasmarshall:


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -2170,6 +2170,9 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
     const error = err as any;
     switch (error.code) {
       case 400:
+        if (error.body?.message.contains('INVALID_EMAIL')) {
```
tomkit, line 2173: Is there a more deterministic way to do this and/or can you point to the response signature from their API?

asdfryan (approved):

tomkit (approved):

PR #1675, opened 2023-09-29
fix: use fieldmappings for common object fetches

lucasmarshall:
Fixes: SUP-581

## Test Plan

Tested locally


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -2094,6 +2094,7 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
     await this.maybeRefreshAccessToken();
 
     const cursor = paginationParams.cursor ? decodeCursor(paginationParams.cursor) : undefined;
+    // TODO should use fieldMappingConfig here
```
asdfryan, line 2097: I think we should just do this IMO.

asdfryan (approved):

PR #1677, opened 2023-10-02
fix: respect schemas in list membership

lucasmarshall:


packages/core/remotes/impl/salesforce/index.ts
```diff
@@ -1476,21 +1476,25 @@ ${modifiedAfter ? `WHERE SystemModstamp > ${modifiedAfter.toISOString()} ORDER B
       case 'contact':
         commonObjectRecords = response.records.map((record) => ({
           ...fromSalesforceContactToContact(record),
+          rawData: toMappedProperties(record, fieldMappingConfig),
```
lucasmarshall, line 1479: We probably should rename raw data since it isn't really raw.

tomkit: Let's chat about the response interface going forward in office

lucasmarshall: > Let's chat about the response interface going forward in office

@tomkit Does that need to happen before we can land this? That discussion would impact all interfaces, not just this one. I'm just following our current convention here.

asdfryan (approved):

PR #1679, opened 2023-10-03
feat: Support phone number during egagement contact creation & retrieval

tonyxiao:
## Test Plan

Manually tested

- [x] Get contact with phone in salesloft (we had a bug, this PR fixed it)
- [x] Create contact with phone in Salesloft
- [x] Create contact with phone in Outreach
- [x] Create contact with phone in Apollo (not officially supported by Apollo in docs, but works emprically)
- [x] Update contact with phone in Apollo

## Future

We should add better typing to `record` and perhaps even runtime validation so this type of error wouldn't occur without being caught.


lucasmarshall (approved):

packages/core/remotes/impl/salesloft/mappers.ts
```diff
@@ -41,7 +41,7 @@ export const fromSalesloftPersonToContact = (record: Record<string, any>): Conta
       postalCode: null,
     },
     emailAddresses: fromSalesloftPersonToEmailAddresses(record),
-    phoneNumbers: fromSalesloftPersonToPhoneNumbers(record.phone_numbers ?? []),
+    phoneNumbers: fromSalesloftPersonToPhoneNumbers(record),
```
asdfryan, line 44: Thanks for the fix! Could you add / modify an existing unit test? (in `salesloft/mappers.test.ts`)

packages/core/remotes/impl/outreach/mappers.ts
```diff
@@ -503,6 +503,10 @@ export const toOutreachProspectPhoneNumbers = (phoneNumbers?: PhoneNumber[]) =>
         case 'work':
           workPhones.push(phoneNumber);
           break;
+        case 'primary':
+          // Defaulting to work phone for primary
```
asdfryan, line 507: `s/work/mobile` ?

Also can you add a test case for this? (and the case where we have both primary + mobile)

tonyxiao: 👍

packages/core/remotes/impl/apollo/mappers.ts
```diff
@@ -190,6 +193,14 @@ export const toApolloContactCreateParams = (params: ContactCreateParams): Record
     email: params.emailAddresses?.[0]?.emailAddress,
     present_raw_address: params.address ? getRawAddressString(params.address) : undefined,
     account_id: params.accountId,
+    // Apollo docs appears to not support phone during creation
+    // However empirically it actually works, so we keep the phone mapping logic here.
```
tomkit, line 197: 🙏 for the comments here

PR #1682, opened 2023-10-03
fix: add customer id to posthog events

lucasmarshall:
Fixes: SUP-657


apps/api/lib/posthog.ts
```diff
@@ -28,6 +28,7 @@ function onResFinished(req: Request, res: Response, err?: any) {
       params: req.params,
       providerName: getProviderNameFromRequest(req),
       applicationId: req.supaglueApplication?.id,
+      customerId: req.customerId,
```
asdfryan, line 31: Since we're doing this already, mind also adding `applicationEnv`?

george-xing (approved):

asdfryan (approved):

PR #1684, opened 2023-10-03
fix: log context for api

tomkit:
- Empirically verified that context.X isn't in the output
- Didn't dig into pino-context code, but used trial-and-error/process of elimination to see under what situations it is logged in the output: using `addLogContext` after all the other middleware seems to work

```
supaglue-api-1          |     context: {
supaglue-api-1          |       "applicationEnv": "production",
supaglue-api-1          |       "applicationName": "My Application",
supaglue-api-1          |       "applicationId": "4acf8358-db1f-4dfa-8c3c-357106afc1a7",
supaglue-api-1          |       "orgId": "e7070cc8-36e7-43e2-81fc-ad57713cf2d3"
supaglue-api-1          |     }
```

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/api/middleware/pino_context.ts
+++ b/apps/api/middleware/pino_context.ts
@@ -1,0 +1,16 @@
+import { configureScope } from '@sentry/node';
+import { addLogContext } from '@supaglue/core/lib/logger';
+import type { NextFunction, Request, Response } from 'express';
+
+export async function pinoAndSentryContextMiddleware(req: Request, res: Response, next: NextFunction) {
+  addLogContext('applicationEnv', req.supaglueApplication.environment);
+  addLogContext('applicationName', req.supaglueApplication.name);
+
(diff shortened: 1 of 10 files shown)


apps/api/middleware/pino_context.ts
```diff
@@ -0,0 +1,16 @@
+import { configureScope } from '@sentry/node';
+import { addLogContext } from '@supaglue/core/lib/logger';
+import type { NextFunction, Request, Response } from 'express';
+
+export async function pinoContextMiddleware(req: Request, res: Response, next: NextFunction) {
```
lucasmarshall, line 5: We should rename it since it's not just pino context we are adding, but also sentry context.

lucasmarshall (changes requested):

tomkit:
addressed comment. merging

PR #1685, opened 2023-10-03
fix: add application id to posthog events

lucasmarshall:
Fixes: SUP-657

### DIFF
--- a/apps/api/lib/posthog.ts
+++ b/apps/api/lib/posthog.ts
@@ -3,13 +3,25 @@ import { getSystemProperties, posthogClient } from '@supaglue/core/lib/posthog';
 import type { NextFunction, Request, Response } from 'express';
 
 function getProviderNameFromRequest(req: Request) {
-  let providerName = req.headers['x-provider-name'] as string;
-
-  if (!providerName && 'state' in req.query && typeof req.query.state === 'string') {
-    ({ providerName } = JSON.parse(req.query.state));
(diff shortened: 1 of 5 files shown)


apps/api/lib/posthog.ts
```diff
@@ -3,13 +3,25 @@ import { getSystemProperties, posthogClient } from '@supaglue/core/lib/posthog';
...
+    return req.headers['x-provider-name'];
+  }
 
-  if (!providerName && 'state' in req.query && typeof req.query.state === 'string') {
-    ({ providerName } = JSON.parse(req.query.state));
+  if ('state' in req.query && typeof req.query.state === 'string') {
+    const { providerName } = JSON.parse(req.query.state);
+    return providerName;
   }
+}
```
tomkit, line 14: It seems like `getProviderNameFromRequest` and `getApplicationIdFromRequest` can return undefined -- when do we expect that to be the case?

Also, when do we expect the providerName to be in the query param instead of the header? Is it authenticated vs unauthenticated? Imo, we should have seperate posthog middleware that's used inside `initRoutes` rather than outside of it

tomkit (approved):
Left a comment and stamping

apps/api/lib/posthog.ts
lucasmarshall: They are in the state for oauth

If these return undefined, the event won't include them. That should be fine as it means that the api call that is being recorded doesn't have them.

PR #1693, opened 2023-10-04
fix: pipedrive incremental date comparison

tomkit:
## Test Plan

Tested this locally

### DIFF
--- a/packages/core/remotes/impl/pipedrive/index.ts
+++ b/packages/core/remotes/impl/pipedrive/index.ts
@@ -801,7 +801,7 @@ export const authConfig: ConnectorAuthConfig = {
 
 function filterForUpdatedAfter<
   R extends {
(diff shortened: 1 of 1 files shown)


packages/core/remotes/impl/pipedrive/index.ts
```diff
@@ -801,9 +801,10 @@ export const authConfig: ConnectorAuthConfig = {
 
 function filterForUpdatedAfter<
   R extends {
-    data: { updated_time?: string }[] | null;
+    data: { update_time?: string }[] | null;
   }
 >(response: R, updatedAfter?: Date): R {
+  console.log('xxx', updatedAfter, response);
```
lucasmarshall, line 807: Remove `console.log`

lucasmarshall (changes requested):

asdfryan (approved):
Nice

PR #1702, opened 2023-10-05
ci: run integration tests

lucasmarshall:


.github/workflows/docker_build_and_push.yml
```diff
@@ -46,7 +50,7 @@ jobs:
     name: Build and Push to Docker Hub
     timeout-minutes: 15
     runs-on:
-      labels: ubuntu-8cores-32gb
+      labels: ubuntu-latest
```
lucasmarshall, line 53: We don't need the bigger machines for this since all the work is done by depot.

.github/workflows/docker_build_and_push.yml
```diff
@@ -117,7 +121,7 @@ jobs:
     name: Deploy to staging
     needs: build
     runs-on:
-      labels: ubuntu-8cores-32gb
+      labels: ubuntu-latest
```
lucasmarshall, line 124: We don't need the bigger machines for this since all this does is hit the k8s API.

.github/workflows/docker_build_and_push.yml
tomkit: There's still some local dependency install and execution that this workflow does itself

I think the bill here is marginal so far so I'd like to err on the side of speed

tomkit (commented):
I think new workflows will run in branches -- can we test that first (we can change the guards for when it should run after it's been tested)

lucasmarshall:
> I think new workflows will run in branches -- can we test that first (we can change the guards for when it should run after it's been tested)

I already tested using [act](https://github.com/nektos/act)

.github/workflows/docker_build_and_push.yml
lucasmarshall: Strongly disagree. We actually don't need the nodejs env or corepack enable here. I'll remove those.

tomkit (approved):

PR #1705, opened 2023-10-06
ci: gate deploys on staging integration tests

lucasmarshall:


scripts/helpers.sh
```diff
@@ -39,3 +39,16 @@ check_checkly_checks() {
         exit 1
     fi
 }
+
+check_github_checks() {
```
tomkit, line 43: Can you add some comments describing what the function is doing?

tomkit (approved):

PR #1706, opened 2023-10-06
fix: apollo sync improvements, paginator stream changes

tomkit:
- Don't retry workflow for `syncObjectRecords` activity to avoid draining api quota
- Apollo is special. Use its own ratelimit function
- Abort retry and fail the workflow on hourly, daily, free plan rate limits
- Increase Apollo page size
- Allow our streams to abort early on certain errors
- Fix Apollo paginating for edge case: no data

## Test Plan

- Synced 202 records while paginating
- Failed on hourly/daily rate limit early

## Deployment instructions

### DIFF
--- a/packages/core/lib/ratelimit.ts
+++ b/packages/core/lib/ratelimit.ts
@@ -3,16 +3,8 @@ import { isAxiosError } from 'axios';
 import { ASYNC_RETRY_OPTIONS, logger } from '.';
 import { TooManyRequestsError } from '../errors';
 
-export const isAxiosRateLimited = (e: any): boolean => {
-  return (
-    isAxiosError(e) &&
-    e.response?.status === 429 &&
-    // Apollo-specific things we shouldn't retry on
-    !e.response.data.contains('This endpoint is only available to paying teams') &&
-    !/^The maximum number of api calls allowed for [a-z0-9/]+ is [0-9]+ times per (hour|day). Please upgrade your plan/.test(
-      e.message
-    )
-  );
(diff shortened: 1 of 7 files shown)


packages/core/lib/ratelimit.ts
```diff
@@ -3,16 +3,8 @@ import { isAxiosError } from 'axios';
 import { ASYNC_RETRY_OPTIONS, logger } from '.';
 import { TooManyRequestsError } from '../errors';
 
-export const isAxiosRateLimited = (e: any): boolean => {
-  return (
-    isAxiosError(e) &&
-    e.response?.status === 429 &&
-    // Apollo-specific things we shouldn't retry on
-    !e.response.data.contains('This endpoint is only available to paying teams') &&
```
tomkit, line 11: `contains` isn't a method on string variable, so this was actually causing the syncs to fail and hang for the max 120 mins

packages/core/lib/ratelimit.ts
```diff
@@ -3,16 +3,8 @@ import { isAxiosError } from 'axios';
...
-    !/^The maximum number of api calls allowed for [a-z0-9/]+ is [0-9]+ times per (hour|day). Please upgrade your plan/.test(
-      e.message
```
tomkit, line 13: the response to match against is in `e.response.data`

asdfryan (approved):
LGTM assuming you also tested this on non-apollo syncs

tomkit:
> LGTM assuming you also tested this on non-apollo syncs

Good call out, ty

PR #1709, opened 2023-10-06
docs: upgrade to docusaurus-openapi-docs v2b4

lucasmarshall:
Also only have current and next versions of docs

Also fix some docs issues

Also remove pub-sub


tomkit (approved):
approved assuming build errors are fixed

PR #1741, opened 2023-10-16
feat: CRM Metadata APIs for custom objects and associations

asdfryan:
This is a re-skin of what albert already implemented elsewhere, with a few changes / simplifications:
- Remove cardinality as we were not actually using it
- Migrate away from entities and use object names instead
- remove `id` from custom object schemas and use `name` instead everywhere
- Supports most fields for hubspot / salesforce (picklists not supported yet)
- Update sidebar

This also adds the open API spec for dealing with association + custom object records, which is unimplemented.
TODO:
- implement endpoints for CRUD custom object records and associations


tomkit:
Listing out some feedback:

- We discussed naming the two types of APIs something like: Custom Object Schemas a Custom Objects (for records)
- There'
Screenshot 2023-10-17 at 10 50 01 AM
s a bit of inconsistency between the casing and spacing of the Custom Object Schemas and Custom Objects APIs

tomkit (commented):
Left comments above

tomkit (approved):
nits:

- Custom Object Schemas and Association Schemas "summary" should use lower case words in the sidebar (they currently differ in documentation convention than Custom Objects and Associations)

PR #1751, opened 2023-10-19
feat: standard object CRUD apis (with cached reads)

asdfryan:
Tested locally

Screenshot 2023-10-19 at 11 41 28 AM
Screenshot 2023-10-19 at 11 41 20 AM
Screenshot 2023-10-19 at 11 41 16 AM


openapi/v2/crm/paths/custom_objects/{object_name}@records@{record_id}.yaml
```diff
@@ -59,12 +59,6 @@ get:
     | Salesforce  | Yes       |
     | Pipedrive   | No        |
     | MS Dynamics | No        |
-  parameters:
```
asdfryan, line 62: bugfix from before. Doesn't make sense to have these params for get By Id

lucasmarshall (approved):

PR #1753, opened 2023-10-19
chore: change response validation errors to warnings to reduce noise

lucasmarshall:
These should be warnings as they flood the logs making it hard to identify true errors that customers are seeing.

### DIFF
--- a/apps/api/middleware/openapi.ts
+++ b/apps/api/middleware/openapi.ts
@@ -41,10 +41,10 @@ export const openapiMiddleware = (specDir: string, version = 'v2') => {
       removeAdditional: true,
     },
     validateResponses: {
-      onError: (error, body, req) => {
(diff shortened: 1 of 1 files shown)


apps/api/middleware/openapi.ts
```diff
@@ -41,10 +41,10 @@ export const openapiMiddleware = (specDir: string, version = 'v2') => {
       removeAdditional: true,
     },
     validateResponses: {
-      onError: (error, body, req) => {
-        logger.error(
+      onError: (err, body, req) => {
+        logger.warn(
           {
-            error,
+            err,
```
lucasmarshall, line 47: pino expects the error to be on the `err` key, not the `error` key

asdfryan (approved):

PR #1759, opened 2023-10-20
fix: hubspot association dedup + custom object sync start

asdfryan:
Tested locally


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -886,13 +886,14 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
         results: response.data.results.map(({ associations, ...rest }) => ({
           ...rest,
           associations: Object.entries(associations ?? {}).reduce((acc, [associatedObjectTypeKey, { results }]) => {
+            const deduped = [...new Set(results.map(({ id }) => id))];
```
tomkit, line 889: nit: `dedupedIds`

tomkit (approved):

PR #1766, opened 2023-10-23
fix: return better error codes for some errors

lucasmarshall:


.github/workflows/production_integration_tests.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     with:
       api-url: https://api.supaglue.io
       customer-id: customer1
-      page-on-failure: true
+      page-on-failure: false
```
lucasmarshall, line 20: Not sure why this is being included. It's already on `main`

PR #1777, opened 2023-10-24
feat: hubspot custom fields api (no picklist)

asdfryan:
Tested locally


tomkit:
I'd like to do a play-through test of this on staging before release

asdfryan:
> I'd like to do a play-through test of this on staging before release

That sounds fine. Could you review this PR so I can land to staging?

packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -1213,6 +1222,86 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
...
+    await retryWhenRateLimited(async () => {
+      await this.maybeRefreshAccessToken();
+      try {
+        await this.#client.crm.properties.groupsApi.create(objectName, {
+          name: groupName,
+          label: groupName,
+        });
+      } catch (e) {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
lucasmarshall, line 1295: Why are we using `@ts-ignore`?

asdfryan: Switched to use any

tomkit (approved):

PR #1791, opened 2023-10-26
chore: support specifying postgres SSL CA certs and client certs/keys

lucasmarshall:
## Test Plan

Tested against a GCP Cloud SQL postgres server configured to require SSL, and also tested with it configured to require client certs.


apps/mgmt-ui/src/components/connectors/destination/PostgresDestinationDetailsPanel.tsx
```diff
@@ -253,15 +269,100 @@ export default function PostgresDestinationDetailsPanel({ isLoading }: PostgresD
...
             }}
             value={sslMode ?? 'disable'}
-            options={[{ value: 'disable' }, { value: 'allow' }, { value: 'prefer' }, { value: 'require' }]}
+            options={[
+              { value: 'disable' },
+              { value: 'no-verify' },
+              { value: 'prefer' },
+              { value: 'require' },
+              { value: 'verify-ca' },
+              { value: 'verify-full' },
```
tomkit, line 283: have we tested all of these combinations? as well as on AWS RDS?

lucasmarshall: Yes, I tested them all. I have not tested RDS, do we have a customer that is needing this there?

lucasmarshall: Tested CA certs on RDS and it works. RDS doesn't support client certs.

tomkit (approved):
Looks like some build issues

PR #1793, opened 2023-10-27
feat: pause sync on certain error types and send webhook, notification

tomkit:
Backend for pausing syncs and notifications:
- Add a layer of inheritance for SGSyncWorker errors. Add `SGConnectionNoLongerAuthenticated` error that will pause syncs and send a webhook+notification
- Add ses for email notifications
- Add `application.email` to our data model
- Add `notificationService`

TODO: data-drive recipient email in mgmt portal

## Test Plan

Tested locally

## Deployment instructions


openapi/v2/mgmt/webhooks/sync.paused.yaml
```diff
@@ -0,0 +1,63 @@
+post:
```
lucasmarshall, line 1: Did we add this to svix already?

packages/core/errors.ts
```diff
@@ -125,18 +125,38 @@ export class TooManyRequestsError extends HTTPError {
...
+  problemType = 'SG_TERMINAL_TOO_MANY_REQUESTS_ERROR';
   constructor(message: string, cause?: Error) {
     super(message, cause);
   }
 }
 
-export class NotModifiedError extends HTTPError {
-  code = 304;
-  problemType = 'NOT_MODIFIED';
+// Throw this when you want run_object_sync to pause after the current execution
```
lucasmarshall, line 157: Should we make the name more general here? like `SGPauseSync`

packages/sync-workflows/activities/pause_sync.ts
```diff
@@ -0,0 +1,65 @@
...
+        connection_id: pausedSync.connectionId,
+        customer_id: connection.customerId,
+        provider_name: connection.providerName,
+        type: 'object',
+        object_type: pausedSync.objectType,
+        object: pausedSync.object,
+      },
+      connection.applicationId,
+      runId
+    );
```
lucasmarshall, line 45: We should emit this webhook on any pause, not just pauses due to errors. So it should be in `syncService.pauseSync` instead.

lucasmarshall (changes requested):

openapi/v2/mgmt/webhooks/sync.paused.yaml
tomkit: Need to import it

packages/core/errors.ts
tomkit: I think the error should reflect why it's being thrown and not the action that will happen

I can't anticipate all of the causes where we want to pause and we will probably want to discriminate between different types of errors, so I'd like to keep it specific for now and we can generate once we get there

packages/core/dependency_container.ts
```diff
@@ -75,6 +78,14 @@ function createCoreDependencyContainer(): CoreDependencyContainer {
   const pgPool = getPgPool(process.env.SUPAGLUE_DATABASE_URL!);
   const systemSettingsService = new SystemSettingsService(prisma);
 
+  const sesClient = new SESv2Client({
+    region: 'us-west-2',
+    credentials: {
+      accessKeyId: process.env.AWS_ACCESS_KEY_ID!,
+      secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY!,
+    },
```
lucasmarshall, line 86: Can't we just add this to the worker role and let STS take care of it?

lucasmarshall (changes requested):

lucasmarshall (approved):

packages/core/dependency_container.ts
tomkit: Yea, can you point me to the cloud repo for where to add it

PR #1796, opened 2023-10-27
fix: custom object syncs + operations

asdfryan:
We now look up the ID of the custom object before performing any API calls on hubspot.

Tested locally


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -1949,9 +1954,36 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
     return response.results.map((object) => ({ id: object.id, name: object.name }));
   }
 
-  public override async getCustomObjectSchema(id: string): Promise<CustomObjectSchema> {
+  #isAlreadyObjectTypeId(nameOrId: string): boolean {
```
tomkit, line 1957: Can you add a comment here with more context about the Hubspot specific context and shape of object type id you're expecting?

tomkit (approved):

packages/core/remotes/impl/hubspot/index.ts
asdfryan: The shape is clear from the regex, no? It's a number followed by a dash followed by another number.

asdfryan: The context is already added in the following function (see the comments). This is just a helper.

PR #1798, opened 2023-10-27
feat: Add batch create capability and implement for Apollo sequence state

tonyxiao:


apps/api/routes/engagement/v2/sequence_state.ts
```diff
@@ -42,12 +42,26 @@ export default function init(app: Router): void {
       req: Request<CreateSequenceStatePathParams, CreateSequenceStateResponse, CreateSequenceStateRequest>,
       res: Response<CreateSequenceStateResponse>
     ) => {
-      const id = await engagementCommonObjectService.create(
-        'sequence_state',
-        req.customerConnection,
-        camelcaseKeysSansCustomFields(req.body.record)
-      );
-      return res.status(201).send({ record: { id } });
+      const [id, ids] = await Promise.all([
```
asdfryan, line 45: I would rather create a separate batch create endpoint for clarity, and maintain consistency with other common APIs and REST-fulness on this endpoint.

tomkit: We discussed in-office just extending the current endpoint and starting a convention of `record` vs `records` for single vs batch records for expediency where the shape of the input is the same between the two since

openapi/v2/engagement/paths/sequence_states.yaml
```diff
@@ -15,8 +19,11 @@ post:
           properties:
             record:
               $ref: ../components/schemas/create_sequence_state.yaml
-          required:
-            - record
+            records:
```
asdfryan, line 22: This should be a `oneOf` between `record` and `records`

apps/api/routes/engagement/v2/sequence_state.ts
asdfryan: I see. In that case, it should be a `oneOf` and we should disallow setting both.

Instead of doing a `Promise.all` we should just condition on whether or not the payload has `record` or `records`. Right now, the code is misleading and making it seem like it's awaiting on both operations.

openapi/v2/engagement/paths/sequence_states.yaml
```diff
@@ -38,6 +45,11 @@ post:
                 $ref: ../../../common/components/schemas/errors.yaml
               record:
                 $ref: ../../../common/components/schemas/created_model.yaml
+              records:
```
asdfryan, line 48: IMO we can just have `records` and get rid of `record` (have it return an array of size 1 in the case there's only 1).

apps/api/routes/engagement/v2/sequence_state.ts
tonyxiao: This is the intended behavior. If user happens to pass in both `record` and `records`, we will create all of them, and return the corresponding IDs in the right place, which I think is fine.

tonyxiao: I think `oneOf` doesn't show as well in the generated openAPI documentation

openapi/v2/engagement/paths/sequence_states.yaml
tonyxiao: Yea i'd prefer that too but we didn't want to introduce any backwards incompatible breaking change.

apps/api/routes/engagement/v2/sequence_state.ts
tomkit: I agree with `oneOf`, if we provide a `title` docusaurus displays it relatively well

Where it falls short is the example section, but we can punt on that

packages/core/services/common_objects/engagement/common_object_service.ts
```diff
@@ -31,36 +31,54 @@ export class EngagementCommonObjectService {
...
+    records: Array<EngagementCommonObjectTypeMap<T>['createParams']>
+  ): Promise<string[]> {
     const [remoteClient, providerName] = await this.#remoteService.getEngagementRemoteClient(connection.id);
 
     const end = remoteDuration.startTimer({ operation: 'create', remote_name: providerName });
-    const res = await remoteClient.createCommonObjectRecord(type, params);
+
+    const resArray = await (remoteClient.batchCreateCommonObjectRecord
+      ? remoteClient.batchCreateCommonObjectRecord(type, records)
+      : Promise.all(records.map((params) => remoteClient.createCommonObjectRecord(type, params))));
```
asdfryan, line 45: I don't think we should pretend to provide a batch endpoint for providers that don't have this capability.

If the first 3 succeeds than the 4th one fails, then the expected behavior is to rollback the first 3. (Or provide information in the response and/or error that tells you which ones were created)

tomkit:
I'll help land - we can address the `oneof` after

packages/core/services/common_objects/engagement/common_object_service.ts
tomkit: +1 on providing information on partial success/failure

tonyxiao: @george-xing do you think we should provide unified experience here for outreach / salesloft which don't support batch sequence state creation?

Also fwiw this is only whitelisted for sequence states, so we won't by default do this for other types.

tomkit:
Agreed with the feedback- let's address as a fast-follow

tomkit (approved):

openapi/v2/engagement/paths/sequence_states.yaml
tonyxiao: Don't have strong enough opinion so updated it to `oneOf`

apps/api/routes/engagement/v2/sequence_state.ts
tonyxiao: cool updated it to use `oneOf`

packages/core/services/common_objects/engagement/common_object_service.ts
tonyxiao: and how should information about partial failure be provided? Do we want to create an `error` object or just return `null` instead of record?

tonyxiao:
Thanks @tomkit and @asdfryan . I added the fast follow PR here https://github.com/supaglue-labs/supaglue/compare/main...tony/apollo-batch-sequence-state

tonyxiao:
https://github.com/supaglue-labs/supaglue/pull/1800

PR #1800, opened 2023-10-27
fix: Improve / bugfix the batch sequence state create endpoint

tonyxiao:
use oneOf
make userId and mailboxId properly optional for apollo


openapi/v2/engagement/paths/sequence_states.yaml
```diff
@@ -15,15 +15,22 @@ post:
...
-          type: object
-          properties:
-            record:
-              $ref: ../components/schemas/create_sequence_state.yaml
-            records:
-              type: array
-              description: Will use the batch endpoints when possible (e.g. Apollo)
-              items:
-                $ref: ../components/schemas/create_sequence_state.yaml
+          oneOf:
```
asdfryan, line 18: We should also update the descriptions / summaries for this endpoint to reflect that this can add multiple sequence states.

asdfryan (commented):
Could you also update the typescript code (in routes)?

openapi/v2/engagement/paths/sequence_states.yaml
tonyxiao: This is already in the endpoint description

tonyxiao: CleanShot 2023-10-27 at 16 44 28@2x

asdfryan: Sorry, I guess I was looking at the response description, which should be updated to `Sequence state(s) created`. Also the summary for this endpoint, but this is minor.

asdfryan (approved):

PR #1834, opened 2023-11-01
chore: pause emit pause reason

asdfryan:


lucasmarshall (changes requested):
I don't see where we are setting the pause reason for manual pauses. Can we add that?

asdfryan:
> I don't see where we are setting the pause reason for manual pauses. Can we add that?

The way @tomkit implemented this, the notification / email path is only triggered for autopaused workflows. Doing this would require a much more significant change. (I agree with you though)

lucasmarshall:
> > I don't see where we are setting the pause reason for manual pauses. Can we add that?
> 
> The way @tomkit implemented this, the notification / email path is only triggered for autopaused workflows. Doing this would require a much more significant change. (I agree with you though)

We do send webhooks on manual pauses though. It would be good to have the reason there. (That's what's powering the slack notifications)

asdfryan:
> > > I don't see where we are setting the pause reason for manual pauses. Can we add that?
> > 
> > 
> > The way @tomkit implemented this, the notification / email path is only triggered for autopaused workflows. Doing this would require a much more significant change. (I agree with you though)
> 
> We do send webhooks on manual pauses though. It would be good to have the reason there. (That's what's powering the slack notifications)

Done.

lucasmarshall (approved):

PR #1836, opened 2023-11-01
chore: autopause for misconfigured entity

asdfryan:


lucasmarshall (changes requested):
Why are we deleting so many docs here?

asdfryan:
> Why are we deleting so many docs here?

Good question. I don't know what happened here other than maybe a bad merge? Let me try again.

lucasmarshall (approved):

packages/sync-workflows/workflows/run_object_sync.ts
```diff
@@ -52,6 +52,9 @@ function getPauseReasonIfShouldPause(err: any): string | undefined {
   if (err.cause?.failure?.message.startsWith('No entity mapping found for entity')) {
     return err.cause.failure.message;
   }
+  if (err.cause?.failure?.message.startsWith('Additional field mappings are not allowed')) {
```
tomkit, line 55: Separate point: Errors we throw should be deterministic and the variable part of the message should be separate from the message

PR #1839, opened 2023-11-01
feat: write custom objects into a single table and create convenience views

tonyxiao:
## Test Plan

Manually tested this works for both custom objects and standard objects in both managed sync and customer postgres settings. Did not implement this for postgres so did not test its implementation. 

  | Standard Objects | Custom Objects
-- | -- | --
Managed sync | Tested ✅ | Tested ✅
Customer postgres | Tested ✅ | Tested ✅
Customer bigQuery | No change | No change

Also tested that that this works for the custom object lists endpoint when using Supaglue managed destination (thanks to the views)

https://github.com/supaglue-labs/supaglue/assets/696842/d1c6060d-82be-47aa-a7a3-ec6cfba6ea0e

### DIFF
--- a/packages/core/destination_writers/postgres_impl.ts
+++ b/packages/core/destination_writers/postgres_impl.ts
@@ -40,6 +40,9 @@ import {
   stripNullCharsFromString,
 } from './util';
 
+export const kCustomObject = 'custom_objects';
+export type ObjectType = 'standard' | 'custom';
+
 export class PostgresDestinationWriterImpl {
   async upsertCommonObjectRecordImpl<P extends ProviderCategory, T extends CommonObjectTypeForCategory<P>>(
     client: PoolClient,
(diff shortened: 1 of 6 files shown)


packages/core/destination_writers/postgres_impl.ts
```diff
@@ -490,15 +504,21 @@ DO UPDATE SET (${columnsToUpdateStr}) = (${excludedColumnsToUpdateStr})`,
         // TODO: This may have performance implications. We should look into this later.
         // https://github.com/supaglue-labs/supaglue/issues/497
         await client.query(`INSERT INTO ${qualifiedTable} (${columns.join(',')})
-SELECT DISTINCT ON (_supaglue_id) ${columns.join(',')} FROM ${dedupedTempTable} OFFSET ${offset} limit ${batchSize}
-ON CONFLICT (_supaglue_application_id, _supaglue_provider_name, _supaglue_customer_id, _supaglue_id)
+SELECT DISTINCT ON (_supaglue_id${objectType === 'custom' ? ', _supaglue_object_name' : ''}) ${columns.join(
```
tomkit, line 507: Thoughts on keeping this SQL interpolation mostly declaratively by making the ternary/conditional output a named function?

tomkit (approved):
It'd be nice to clean up some of the SQL-generation via interpolation where ternary operators are used so it reads more clearly what SQL is generated

tonyxiao:
Clean up a little of the inline interpolations by extracting into a variable, but it's not obvious how to make it significantly better.

PR #1841, opened 2023-11-02
chore: script to generate sync runs

tomkit:
## Test Plan

[Describe test plan here]

## Deployment instructions


asdfryan (approved):

lucasmarshall (commented):
Why do we need this?

PR #1843, opened 2023-11-02
feat: Sequence integration tests

tonyxiao:
Fixes: #### 

[Describe your change here]

## Test Plan

[Describe test plan here]

## Deployment instructions

### DIFF
--- a/apps/api/routes/engagement/v2/sequence.integration.test.ts
+++ b/apps/api/routes/engagement/v2/sequence.integration.test.ts
@@ -1,7 +1,7 @@
 /**
- * Tests accounts endpoints
+ * Tests sequences endpoints
  *
- * @group integration/engagement/v2/accounts
+ * @group integration/engagement/v2/sequences
  * @jest-environment ./integration-test-environment
  */
 
@@ -16,7 +16,7 @@ describe('sequence', () => {
 
   beforeEach(() => {
     testSequence = {
-      name: `Sequence ${Math.random().toString()}`,
+      name: `Test Sequence ${Math.random().toString()}`,
       type: 'team',
       steps: [
         {
@@ -39,7 +39,7 @@ describe('sequence', () => {
           },
         },
         {
-          interval_seconds: 90000,
+          interval_seconds: 172800,
           is_reply: true,
           type: 'call',
           task_note: 'my call notes',
@@ -50,20 +50,9 @@ describe('sequence', () => {
           },
         },
         {
-          interval_seconds: 2160,
+          interval_seconds: 0,
           is_reply: true,
           type: 'task',
-          task_note: 'my task notes',
-          template: {
-            name: 'what is up~!',
-            subject: 'hello world again',
-            body: 'Hi there, how are you doing?',
-          },
-        },
-        {
-          interval_seconds: 30,
-          is_reply: true,
-          type: 'linkedin_send_message',
           task_note: 'my task notes',
           template: {
             name: 'what is up~!',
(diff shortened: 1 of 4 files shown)


asdfryan (changes requested):
I think we should also test `sequenceState`.

Something new that we decided to commit to was to also test invalid request bodies and test that the error responses include relevant context.

packages/core/remotes/impl/salesloft/mappers.ts
```diff
@@ -396,9 +396,7 @@ export const toSalesloftCadenceStepImportParams = (step: SequenceStepCreateParam
           day,
           automated: step.type === 'auto_email',
           automated_settings:
-            step.type === 'auto_email' && delayInMins !== 0
-              ? { send_type: 'after_time_delay', delay_time: delayInMins }
-              : undefined,
+            step.type === 'auto_email' ? { send_type: 'after_time_delay', delay_time: delayInMins } : undefined,
```
asdfryan, line 399: I think you may have to update the existing mapper unit tests

apps/api/routes/engagement/v2/sequence.integration.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        {
+          headers: { 'x-provider-name': providerName },
+        }
+      );
+      expect(response.status).toEqual(201);
+      expect(response.data.record?.id).toBeTruthy();
+      addedObjects.push({
+        id: response.data.record?.id as string,
+        providerName,
+        objectName: 'account',
```
asdfryan, line 81: This doesn't look quite right? It should be `sequence` -- and there should probably be additional cleanup code written in `integration-test-environment.js`

tonyxiao: Thanks for the catch!

tonyxiao:
Done for both

apps/api/routes/engagement/v2/sequence.integration.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      const response = await apiClient.post<CreateSequenceResponse>(
+        '/engagement/v2/sequences',
+        { record: testSequence },
+        {
+          headers: { 'x-provider-name': providerName },
+        }
+      );
+      expect(response.status).toEqual(201);
+      expect(response.data.record?.id).toBeTruthy();
+      addedObjects.push({
```
asdfryan, line 115: Does this get cleaned up at the end even if there are active contacts on it?

apps/api/routes/engagement/v2/sequence.integration.test.ts
```diff
@@ -0,0 +1,177 @@
...
+const APOLLO_MAILBOX_ID = '651c4cb3520e8800a3808e8a';
+
+describe('sequence', () => {
+  let testSequence: ReturnType<typeof getTestSequence>;
+
+  beforeEach(() => {
+    testSequence = getTestSequence();
+  });
+
+  test.skip(`Error body`, async () => {
```
asdfryan, line 93: are we intentionally skipping?

asdfryan (approved):

apps/api/routes/engagement/v2/sequence.integration.test.ts
tonyxiao: image
Yes, it appears so.

tonyxiao: Though delete is only supported in outreach api though.

tonyxiao: image


And for apollo I don't think there is delete at all, only archive.

PR #1844, opened 2023-11-02
fix: missing partitions for contacts and sequence states

asdfryan:
This was the only missing piece to make managed syncs work for engagement.


tomkit (approved):
Similar comment here: it would be nice to have the interpolated SQL generation read more declaratively

PR #1852, opened 2023-11-03
fix: respect customer id in path for connection apis. peg integration tests to release tag for prod

tomkit:
## Test Plan

- Tested negative and positive cases for LIST, GET, _provider_user_id

## Deployment instructions

### DIFF
--- a/apps/api/routes/mgmt/v2/customer/connection/connection.integration.test.ts
+++ b/apps/api/routes/mgmt/v2/customer/connection/connection.integration.test.ts
@@ -1,0 +1,60 @@
+/**
+ * Tests connection endpoints
+ *
+ * @group integration/mgmt/v2/customers/connections
+ * @jest-environment ./integration-test-environment
+ */
+
+import type { GetConnectionResponse, GetConnectionsResponse } from '@supaglue/schemas/v2/mgmt';
+
+describe('connection', () => {
+  test(`LIST (200) then GET (200)`, async () => {
+    const CUSTOMER_ID = 'customer1';
+    const connectionsResponse = await apiClient.get<GetConnectionsResponse>(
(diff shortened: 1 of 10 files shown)


apps/api/routes/mgmt/v2/customer/connection/index.ts
```diff
@@ -29,6 +30,9 @@ const { connectionService, connectionAndSyncService, remoteService } = getDepend
 export default function init(app: Router): void {
   const connectionRouter = Router({ mergeParams: true });
 
+  connectionRouter.use(customerPathMiddleware);
+  // connectionRouter.use(pinoAndSentryContextMiddleware);
```
asdfryan, line 34: Remove?

asdfryan (approved):

asdfryan (changes requested):
Add integration tests for this use case?

apps/api/routes/mgmt/v2/customer/connection/connection.integration.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  test(`LIST (404)`, async () => {
+    const CUSTOMER_ID = 'nonexistentcustomer';
+    const connectionsResponse = await apiClient.get<GetConnectionsResponse>(
+      `/mgmt/v2/customers/${CUSTOMER_ID}/connections`
+    );
+    expect(connectionsResponse.status).toEqual(404);
+  });
+
```
asdfryan, line 32: I think the scenario this PR is detecting is the case where both the `connectionId` and `customerId` exists, but the `connectionId` does not belong to the `customerId`, right? Can we add a test case for that?

asdfryan (approved):

PR #1855, opened 2023-11-03
fix: Apollo docs and handle contact already added to sequence

tonyxiao:
Fixes: #### 

[Describe your change here]

## Test Plan

[Describe test plan here]

## Deployment instructions


tonyxiao:
ok @asdfryan handled according to our discussion.

asdfryan (commented):
Add integration test for this case?

packages/core/remotes/impl/apollo/index.ts
```diff
@@ -487,33 +487,41 @@ class ApolloClient extends AbstractEngagementRemoteClient {
...
+    return await Promise.all(
+      records.map(async (record) => {
+        let contact = res.contacts.find((c) => c.id === record.contactId);
+        if (!contact) {
+          // Handle situation where contact have already been added to the sequence. However still accounting for
+          // other errors where contact could not be added to sequence
+          contact = await this.#api.getContact({ params: { id: record.contactId } }).then((r) => r.contact);
+        }
+        if (!contact) {
+          throw new Error(`Unable to find contact ${record.contactId} in Apollo`);
```
asdfryan, line 499: Should this be a 400 or a 500?

If 400, we should use `BadRequestError`. If 500, we should use `InternalServerError`

packages/core/remotes/impl/apollo/index.ts
```diff
@@ -487,33 +487,41 @@ class ApolloClient extends AbstractEngagementRemoteClient {
...
-        record: fromApolloContactCampaignStatusToSequenceState(contact.id, campaignStatus as any),
-      };
-    });
+    return await Promise.all(
+      records.map(async (record) => {
+        let contact = res.contacts.find((c) => c.id === record.contactId);
+        if (!contact) {
+          // Handle situation where contact have already been added to the sequence. However still accounting for
+          // other errors where contact could not be added to sequence
+          contact = await this.#api.getContact({ params: { id: record.contactId } }).then((r) => r.contact);
```
asdfryan, line 496: It's not sufficient to fetch the contact. We should also check that the contact's sequence status contains the sequence for which the request is trying to add.

asdfryan (changes requested):
Some comments

packages/core/remotes/impl/apollo/index.ts
tonyxiao: The code right underneath this line does exactly that and otherwise throw.

packages/core/remotes/impl/apollo/index.ts
tonyxiao: Not sure, it feels like an apollo error, so which one does it fall under?

asdfryan: I actually feel like it should really be a 404 in this case. The customer specified a contactId that doesn't exist.

packages/core/remotes/impl/apollo/index.ts
asdfryan: Oops, I missed that. Sorry

packages/core/remotes/impl/apollo/index.ts
```diff
@@ -496,7 +496,7 @@ class ApolloClient extends AbstractEngagementRemoteClient {
           contact = await this.#api.getContact({ params: { id: record.contactId } }).then((r) => r.contact);
         }
         if (!contact) {
-          throw new Error(`Unable to find contact ${record.contactId} in Apollo`);
+          throw new BadRequestError(`Unable to find contact ${record.contactId} in Apollo`);
```
asdfryan, line 499: Should be `NotFoundError`

asdfryan (approved):

PR #1858, opened 2023-11-03
fix!: custom object records integration tests

asdfryan:
This also includes some bugfixes on the relevant code:
- PATCH should return 200 not 201
- Missing required fields in hubspot should return 400 not 500

### DIFF
--- a/apps/api/routes/crm/v2/custom_objects/custom_object_records.integration.test.ts
+++ b/apps/api/routes/crm/v2/custom_objects/custom_object_records.integration.test.ts
@@ -1,0 +1,230 @@
+/**
+ * Tests custom object records endpoints
+ *
+ * @group integration/crm/v2/custom_objects
+ * @jest-environment ./integration-test-environment
+ */
+
+import type {
+  CreateCustomObjectRecordResponse,
+  GetCustomObjectRecordResponse,
+  UpdateCustomObjectRecordResponse,
+} from '@supaglue/schemas/v2/crm';
+
+export const PERMANENT_CUSTOM_OBJECT_NAME = 'PermanentCustomObject';
+
+type PermanentCustomObject = {
+  name: string;
+  description__c: string;
+  int__c?: number;
+  double__c?: number;
+  bool__c?: boolean;
+};
+
+// Permanent custom object schema:
+// {
+//       "name": "PermanentCustomObject",
+//       "description": "Permanent Custom Object to be used for integration tests",
+//       "labels": {
+//           "singular": "PermanentCustomObject",
+//           "plural": "PermanentCustomObjects"
+//       },
+//       "primary_field_id": "name",
+//       "fields": [
+//           {
+//               "label": "My Name Field",
+//               "id": "name",
+//               "is_required": true,
+//               "type": "text"
+//           },
+//           {
+//               "label": "My Description Field",
+//               "id": "description__c",
+//               "is_required": true,
+//               "type": "textarea"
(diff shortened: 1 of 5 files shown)


apps/api/routes/crm/v2/custom_objects/custom_object_records.integration.test.ts
```diff
@@ -0,0 +1,230 @@
...
+      description__c: 'my desc here',
+      int__c: 1,
+      double__c: 0.1,
+      bool__c: true,
+    };
+  });
+
+  // describe.each(['hubspot'])('%s', (providerName) => {
+  describe.each(['hubspot', 'salesforce'])('%s', (providerName) => {
+    test(`Post /`, async () => {
```
tomkit, line 87: this is testing quite a bit of things: can you make the test more descriptive or annotate in the test what mutations you're making and what you're testing for?

packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -2315,7 +2315,10 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
 
     switch (code) {
       case 400:
-        if (message === 'one or more associations are not valid') {
+        if (
+          message === 'one or more associations are not valid' ||
+          message.includes('Some required properties were not set')
```
tomkit, line 2320: `startsWith` or `includes`?

tomkit:
This should be a `fix!` PR not a `chore`

packages/core/remotes/impl/hubspot/index.ts
asdfryan: it says `includes` here.

apps/api/routes/crm/v2/custom_objects/custom_object_records.integration.test.ts
asdfryan: I am using the same pattern / precedent as all the other integration tests.

asdfryan:
> This should be a `fix!` PR not a `chore`

There are no breaking changes in this PR.

apps/api/routes/crm/v2/custom_objects/custom_object_records.integration.test.ts
tomkit: Can we improve upon it for readability; I haven't gotten to reviewing every PR so some patterns we can improve upon

tomkit:
> > This should be a `fix!` PR not a `chore`
> 
> There are no breaking changes in this PR.

The response code change is breaking

packages/core/remotes/impl/hubspot/index.ts
tomkit: I'm more asking what it should be: it looks like we're expecting it to be the start of the message, and the end of the message is dynamic?

asdfryan: If you look at the integration test I added, you can see that the full error message is `Error creating {{CUSTOM_OBJ_NAME}}. Some required properties were not set.`

apps/api/routes/crm/v2/custom_objects/custom_object_records.integration.test.ts
asdfryan: In that case I would like to open a separate PR where I can address this holistically across all of our integration tests. Does that sound good to you?

tomkit (approved):

PR #1860, opened 2023-11-04
fix: delete connection

asdfryan:


tonyxiao (approved):

scripts/prepare_release.sh
```diff
@@ -13,14 +13,14 @@ OLD_VERSION=$(jq -r .version ./package.json)
 VERSION=$1
 
 # fail if not clean working directory
-# if [ -n "$(git status --porcelain)" ]; then
-#   echo "Working directory not clean. Please stash all changes before running this script."
-#   exit 1
-# fi
```
lucasmarshall, line 19: Why was this commented out and why is uncommenting it included here?

scripts/prepare_release.sh
```diff
@@ -13,14 +13,14 @@ OLD_VERSION=$(jq -r .version ./package.json)
...
+if [ -n "$(git status --porcelain)" ]; then
+  echo "Working directory not clean. Please stash all changes before running this script."
+  exit 1
+fi
 
 source $DIR/helpers.sh
 check_checkly_checks
-# check_github_checks
```
lucasmarshall, line 23: Same as above

tomkit (approved):

scripts/prepare_release.sh
tomkit: We needed to bypass integration tests on Friday to deploy so it was commented out as part of the release

Just piggy-backing on it to get it restored here

todo: let's add a flag for bypassing integration tests for deploys

PR #1863, opened 2023-11-06
chore: switch Lekko SDK usage from server props to API route

DavidSGK:
Fixes: #### 

- Upgraded Lekko node server SDK: 0.0.9 -> 0.2.0
- Moved Lekko config fetches from getServerSideProps (which caused delays on each page load, see #1840) to an API route
    - API route is at /api/internal/lekko and currently fetches all configs
    - Cleaned up to remove home CTA button configs which seem to not be used anymore
- Added hook for querying new Lekko API route
- Updated relevant components to use new hook

## Test Plan

- Tested locally

## Deployment instructions


tomkit:
Thanks @DavidSGK !

tomkit (approved):

PR #1864, opened 2023-11-06
feat!: CRM list custom objects endpoint should also include labels

asdfryan:
Tested locally

This contains a breaking interface change

Screenshot 2023-11-06 at 2 34 18 PM


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -1952,12 +1953,24 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
     return HUBSPOT_STANDARD_OBJECT_TYPES as unknown as string[];
   }
 
-  public override async listCustomObjectSchemas(): Promise<SimpleCustomObjectSchema[]> {
+  public override async listCustomObjectSchemasDeprecated(): Promise<SimpleCustomObjectSchemaDeprecated[]> {
```
tomkit, line 1956: Can you annotate in tsdoc too: 
```
/**
 * @deprecated
 */

tomkit (commented):
Is there a staging env I can play-through this on?

tomkit (approved):
Approving with some comments

asdfryan:
> Is there a staging env I can play-through this on?

Yes -- any of them that has hubspot connected.

PR #1866, opened 2023-11-06
fix: contact update + associations for single GETs

asdfryan:
This PR fixes 2 bugs:
- PATCH for `contact` when only providing the `account_id`
- property populate associations for single `GET` endpoints

Tested locally + added integration test

Screenshot 2023-11-06 at 3 15 47 PM


apps/api/routes/crm/v2/contact.integration.test.ts
```diff
@@ -119,6 +120,81 @@ describe('contact', () => {
...
+        {
+          headers: { 'x-provider-name': providerName },
+        }
+      );
+
+      expect(updateResponse.status).toEqual(200);
+
+      // Pipedrive does not have read after write guarantees, so we need to wait
+      if (providerName === 'pipedrive') {
+        await new Promise((resolve) => setTimeout(resolve, 12000));
```
tomkit, line 184: were we waiting 30s in other places?

asdfryan: I can change it to 30 sure.

packages/core/remotes/impl/hubspot/mappers.ts
```diff
@@ -127,6 +127,9 @@ export const fromHubSpotContactToContact_v2 = (hubspotSimplePublicObject: Record
   if (associations?.company?.length) {
     accountId = associations.company[0] ?? null;
   }
+  if (associations?.companies?.length) {
+    accountId = associations.companies[0] ?? null;
```
tomkit, line 131: Separate point: Should we be defining our default association id to have more provider-scoped semantics? E.g. for hubspot we would filter for primary company

tomkit (approved):

tomkit:
Helping to land to test. Let's fix-forward for any issues that arise.

packages/core/remotes/impl/hubspot/mappers.ts
asdfryan: Actually this is an old version of the code that should be deleted. We shouldn't have to do this anymore. I'll make a PR to revert these.

PR #1878, opened 2023-11-08
feat: support account hubspot lists and migrate to v3 API

asdfryan:
Tested locally

Screenshot 2023-11-08 at 3 17 25 PM

image


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -2225,41 +2225,57 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
...
+          count: pageSize,
+          offset,
+        },
+        {
+          headers: {
+            Authorization: `Bearer ${this.#config.accessToken}`,
+          },
+        }
+      );
+      const objectTypeId = objectType === 'contact' ? '0-1' : '0-2';
```
tomkit, line 2255: What do `0-1` and `0-2` mean? Can you annotate?

tomkit (commented):
I think we need to implement listlistMembership for accounts (and migrate that) too

lucasmarshall (approved):

tomkit (approved):
Let's do a play-through for this

PR #1879, opened 2023-11-09
chore: dedupe in first stream step for perf improvement

tomkit:
- push deduping to first stream instead of at DB query time into final table

## Test Plan

- Tested against Hubspot with Supaglue common and standard objects
- Tested archiving, unarchiving and setting the page limit to 1 so archive/unarchive events happen across pages

## Deployment instructions

### DIFF
--- a/packages/core/destination_writers/postgres_impl.ts
+++ b/packages/core/destination_writers/postgres_impl.ts
@@ -132,7 +132,6 @@ DO UPDATE SET (${columnsToUpdateStr}) = (${excludedColumnsToUpdateStr})`,
     const childLogger = logger.child({ connectionId, providerName, customerId, commonObjectType });
     const qualifiedTable = `"${schema}".${table}`;
     const tempTable = `temp_${table}`;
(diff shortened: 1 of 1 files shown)


asdfryan (approved):

packages/core/destination_writers/postgres_impl.ts
```diff
@@ -235,13 +244,8 @@ DO UPDATE SET (${columnsToUpdateStr}) = (${excludedColumnsToUpdateStr})`,
       const batchSize = 10000;
       for (let offset = 0; offset < tempTableRowCount; offset += batchSize) {
         childLogger.info({ offset }, 'Copying from deduped temp table to main table [IN PROGRESS]');
-        // IMPORTANT: we need to use DISTINCT ON because we may have multiple records with the same id
-        // For example, hubspot will return the same record twice when querying for `archived: true` if
-        // the record was archived, restored, and archived again.
```
tomkit, line 240: I did not observe this for Hubspot

PR #1886, opened 2023-11-10
fix: heartbeating, other temporal issues

lucasmarshall:


packages/core/lib/apollo_ratelimit.ts
```diff
@@ -50,6 +50,7 @@ export const retryWhenAxiosApolloRateLimited = async <Args extends any[], Return
     factor: 2,
     // create some jitter so concurrent apollo syncs can make progress
     randomize: true,
-    minTimeout: 2000,
+    minTimeout: 2_000,
+    maxTimeout: 60_000,
```
tomkit, line 54: why do we need to add a max here?

asdfryan: I'm assuming it's because if it goes unbounded, it will cause the heartbeat to time out.

lucasmarshall: Actually I guess it's not needed anymore since we always heartbeat every 10 sec.

asdfryan (approved):

tomkit (approved):

PR #1891, opened 2023-11-10
chore: simplify search interface + add integration tests

asdfryan:
## Before
```
{
    "filter": {
        "key": "email",
        "value": "my@email.com"
    }    
}
```

## After
```
{
    "filter": {
        "email": "my@email.com"
    }    
}
```


apps/api/routes/crm/v2/contact.integration.test.ts
```diff
@@ -265,5 +266,48 @@ describe('contact', () => {
...
       // expect(dbContact2.rows[0].addresses).toEqual(testContact.addresses);
     }, 120_000);
+
+    // Search only supported for hubspot and salesforce
+    testIf(
+      ['salesforce', 'hubspot'].includes(providerName),
+      `Test that POST followed by SEARCH has correct data`,
+      async () => {
+        const email = `me+${Math.random()}@example.com`;
+        const response = await apiClient.post<CreateContactResponse>(
```
tomkit, line 276: Tony's client changes should be in: let's use that client going forward?

tomkit (approved):

apps/api/routes/crm/v2/contact.integration.test.ts
asdfryan: will do in a separate PR. Seems like the right place to make this change is at the `apiClient` layer.

PR #1893, opened 2023-11-11
feat: engagement sequence state search

asdfryan:
This PR also adds Get contact by ID for Apollo via private API, which was previously disabled. This is needed to be able to filter by contact Id in sequence state.

Tested locally + added new integration tests
image


openapi/v2/engagement/openapi.bundle.json
```diff
@@ -1229,6 +1229,95 @@
         }
       ]
     },
+    "/sequence_states/_search": {
```
tomkit, line 1232: If we think we'll be doing cached vs uncached, let's use the same flag we have for list endpoints

openapi/v2/engagement/paths/sequence_states@search.yaml
```diff
@@ -0,0 +1,59 @@
...
+    | Salesloft | contact_id, sequence_id |
+    | Outreach  | contact_id, sequence_id |
+
+  tags:
+    - Sequence States
+  security:
+    - x-api-key: []
+  parameters:
+    - $ref: ../../../common/components/parameters/query/include_raw_data.yaml
+    - $ref: ../../../common/components/parameters/query/page_size.yaml
```
tomkit, line 20: We should use `remote_provider_page_size` for remote and `page_size` for cached

tomkit: I.e. we shouldn't allow users to specify 1000 page size against remotes

tomkit (approved):

openapi/v2/engagement/openapi.bundle.json
asdfryan: Done

PR #1907, opened 2023-11-14
chore: more autopauses

asdfryan:
Add autopauses for:
- revoked Apollo API Keys
- getaddrinfo ENOTFOUND for Microsoft
- resource not found for Salesforce (indicates permission issue)

This also fixes error handling for Microsoft in general

### DIFF
--- a/packages/core/remotes/impl/ms_dynamics_365_sales/index.ts
+++ b/packages/core/remotes/impl/ms_dynamics_365_sales/index.ts
@@ -33,6 +33,7 @@ import {
   NotFoundError,
   NotModifiedError,
   RemoteProviderError,
+  SGConnectionNoLongerAuthenticatedError,
   UnauthorizedError,
 } from '../../../errors';
 import type { ConnectorAuthConfig } from '../../base';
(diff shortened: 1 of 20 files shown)


packages/core/remotes/impl/ms_dynamics_365_sales/index.ts
```diff
@@ -483,21 +488,21 @@ class MsDynamics365Sales extends AbstractCrmRemoteClient {
     ]);
   }
 
-  public override handleErr(err: unknown): unknown {
+  public override async handleErr(err: unknown): Promise<unknown> {
```
tomkit, line 491: Could we await the json outside of this handler and keep it sync?

asdfryan: No because we need the cause

lucasmarshall (approved):

packages/core/remotes/impl/ms_dynamics_365_sales/index.ts
```diff
@@ -349,22 +350,26 @@ class MsDynamics365Sales extends AbstractCrmRemoteClient {
...
+              $filter: updatedAfter ? `modifiedon gt ${updatedAfter?.toISOString()}` : undefined,
+              $expand: expand,
+            }),
+          { headers: this.#headers }
+        );
+        if (heartbeat) {
+          heartbeat();
+        }
+        if (!response.ok) {
+          throw await this.handleErr(response);
```
tomkit, line 367: Would this work?
```
const responseJson = await response.json();
throw await this.handleErr(responseJson);
```

asdfryan: I see. Sure I can make this change.

tomkit (approved):
Unblocking

PR #1930, opened 2023-11-15
docs: clarify force_delete_syncs documentation

asdfryan:
image


openapi/v2/mgmt/paths/sync_configs@{sync_config_id}.yaml
```diff
@@ -23,6 +23,7 @@ put:
       required: false
       schema: 
         type: boolean
+      description: If true, any syncs for any deleted objects will be cascadingly deleted for all customers with this sync config
```
tomkit, line 26: Can we make this a :::danger admonition

tomkit:
Also, there are a few places where the default is to do a cascade delete, can we add a :::danger admonition for those endpoints too

asdfryan:
> Also, there are a few places where the default is to do a cascade delete, can we add a :::danger admonition for those endpoints too

where else do we do a cascade delete?

openapi/v2/mgmt/paths/sync_configs@{sync_config_id}.yaml
asdfryan: apparently not. but I added it to the top level description

lucasmarshall (approved):

PR #1934, opened 2023-11-16
fix: token refresh for outreach + autopause for other providers

asdfryan:


packages/core/remotes/impl/outreach/index.ts
```diff
@@ -1413,7 +1413,7 @@ class OutreachClient extends AbstractEngagementRemoteClient {
             client_id: this.#credentials.clientId,
             client_secret: this.#credentials.clientSecret,
             grant_type: 'refresh_token',
-            refresh_token: 'asdf',
+            refresh_token: this.#credentials.refreshToken,
```
asdfryan, line 1416: This is the fix. I forgot to undo the change after testing.

tomkit (approved):

PR #1935, opened 2023-11-16
chore: autopause hubspot refresh token fail

asdfryan:


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -2450,6 +2457,9 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
         ) {
           return new BadRequestError(message, error);
         }
+        if (message.includes('missing or invalid refresh token')) {
```
asdfryan, line 2460: This is to catch token refreshes coming directly from the hubspot SDK

tomkit (approved):

PR #1938, opened 2023-11-16
feat: New openapi powered clients for interacting with integrations

tonyxiao:
And add a new openapi-fetch powered client


apps/api/index.ts
```diff
@@ -15,6 +15,10 @@ app.set('trust proxy', true);
 const metricsApp = express();
 const port = process.env.SUPAGLUE_API_PORT ? parseInt(process.env.SUPAGLUE_API_PORT) : 8080;
 
+if (process.env.GLOBAL_AGENT_HTTP_PROXY) {
```
asdfryan, line 18: Do we need to add this to the docker-compose?

asdfryan (commented):
Does this maintain all of the existing error handling (e.g. autopauses etc)?

tomkit (approved):
minor comment: can we name generated files '.gen.' too for clarity?

tonyxiao:
I'll add both of those in a follow up PR. 👍

apps/api/index.ts
tonyxiao: Added it to docker-compose file already , for local debugging

PR #1939, opened 2023-11-16
feat: uncached LIST endpoints for Hubspot + Salesforce

asdfryan:


packages/core/remotes/impl/hubspot/index.ts
```diff
@@ -1858,7 +2026,21 @@ class HubSpotClient extends AbstractCrmRemoteClient implements MarketingAutomati
     return id;
   }
 
-  public async listLeads(fieldMappingConfig: FieldMappingConfig, updatedAfter?: Date): Promise<Readable> {
+  public async listLeads(
+    params: CrmListParams,
+    fieldMappingConfig: FieldMappingConfig
+  ): Promise<PaginatedSupaglueRecords<Lead>> {
+    return {
```
tomkit, line 2033: I think we should return an error for this instead of an empty response since this is a developer-facing API. 

`streamLeads` is not directly developer-facing as an interface b/c it just writes to postgres

asdfryan: Done.

tomkit (approved):

PR #1940, opened 2023-11-16
chore: unified error interface

lucasmarshall:
Need to do more testing here, but wanted to get it in  front of people


asdfryan (commented):
This is worth integration testing IMO.

asdfryan (approved):

PR #1943, opened 2023-11-17
fix: create/get common objects for MS dynamics

lucasmarshall:


tonyxiao (approved):

packages/core/remotes/impl/ms_dynamics_365_sales/index.ts
```diff
@@ -234,16 +234,21 @@ class MsDynamics365Sales extends AbstractCrmRemoteClient {
   ): Promise<string> {
     await this.maybeRefreshAccessToken();
     let objectName = plural(object.name);
+    // need our own odata client here as we need to set the `Prefer` header
```
tomkit, line 237: Can you add color on the significance of this header

asdfryan (commented):
We should enable integration tests for dynamics (can do in another PR).

asdfryan (approved):

packages/core/remotes/impl/ms_dynamics_365_sales/index.ts
lucasmarshall: Done

PR #1944, opened 2023-11-17
fix: only apply user middleware to routes where we have a user

lucasmarshall:


asdfryan (approved):

apps/api/middleware/bearer.ts
```diff
@@ -0,0 +1,34 @@
+import { UnauthorizedError } from '@supaglue/core/errors';
+import { addLogContext } from '@supaglue/core/lib/logger';
+import type { NextFunction, Request, Response } from 'express';
+import { verify } from 'jsonwebtoken';
+import jwksClient from 'jwks-rsa';
+
+export async function bearerMiddleware(req: Request, res: Response, next: NextFunction) {
```
asdfryan, line 7: Nit: instead of bearerMiddleware (we also use bearer tokens for our API Key I think), maybe name it something clerk specific?

lucasmarshall: Sure, we can do that, athough are API Key isn't a bearer token. We carry that in the  `x-api-key` header

lucasmarshall: done

PR #1956, opened 2023-11-20
fix: retry on 429 or 5xx failures for native hubspot client. retry on

tomkit:
429 for axios hubspot client

- add axios retry for some places that was missing it
- turn on retries for hubspot client


lucasmarshall (approved):

packages/core/lib/constants.ts
```diff
@@ -20,7 +20,7 @@ export const SCHEMAS_OR_ENTITIES_APPLICATION_IDS = [
 ];
 
 export const ASYNC_RETRY_OPTIONS = {
-  retries: 5,
```
tomkit, line 23: 5 is excessive

PR #1958, opened 2023-11-21
fix: increase time between rate-limited retries for salesloft

lucasmarshall:
We have a customer who is constantly having syncs fail due to salesloft rate limit. This is an attempt to help with that.


supaglue-user (approved):

supaglue-user:
Was the report.html useful by chance?

tonyxiao (approved):
LGTM. though was the report.html useful? accidentally committed before?

lucasmarshall:
> LGTM. though was the report.html useful? accidentally committed before?

Accidentally committed.

PR #1985, opened 2023-11-28
feat: MS Dynamics 365 sales standard object list endpoint

lucasmarshall:


packages/utils/constants.ts
```diff
@@ -1178,4 +1178,1272 @@ export const PROVIDERS_THAT_SUPPORT_SCHEMAS = ['salesforce', 'hubspot', 'pipedri
 
 export const LINEAR_STANDARD_OBJECTS = ['issues', 'comments', 'users', 'projects', 'teams'] as const;
 
+export const MS_DYNAMICS_365_SALES_STANDARD_OBJECTS = [
```
tonyxiao, line 1181: What's the source of this? And how do we keep it up to date when MSDynamic updates?

tonyxiao (approved):
LGTM, non blocking comment

george-xing:
@lucasmarshall could we update the provider docs to point to the same link for the object list? https://docs.supaglue.com/providers/ms_dynamics_365_sales#supported-standard-objects

PR #1986, opened 2023-11-28
fix: restore salesloft retry config

lucasmarshall:
This was removed in #1938 for some reason

### DIFF
--- a/packages/core/remotes/impl/salesloft/index.ts
+++ b/packages/core/remotes/impl/salesloft/index.ts
@@ -186,26 +186,30 @@ class SalesloftClient extends AbstractEngagementRemoteClient {
     heartbeat?: () => void
   ): (next?: string) => Promise<SalesloftPaginatedRecords> {
     return async (next?: string) => {
-      return await retryWhenAxiosRateLimited(async () => {
-        if (heartbeat) {
-          heartbeat();
-        }
-        await this.maybeRefreshAccessToken();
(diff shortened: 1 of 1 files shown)


packages/core/remotes/impl/salesloft/index.ts
```diff
@@ -186,26 +186,30 @@ class SalesloftClient extends AbstractEngagementRemoteClient {
...
+                }
+              : {
+                  ...DEFAULT_LIST_PARAMS,
+                  page: next ? parseInt(next) : undefined,
+                },
+            headers: this.#headers,
+          });
+          return response.data;
+        },
+        // the rate limit is 600/minute shared among all users of the API, so we should wait longer than the usual 1s just to be safe
```
tomkit, line 210: is this only for this endpoint or all salesloft endpoints?

if for all, we should create a rate_limit.ts just for salesloft

tomkit (approved):

packages/core/remotes/impl/salesloft/index.ts
lucasmarshall: Just for this

tomkit: Got it - can we call that out that it's just for this endpoint?

tonyxiao:
Shoot I think this is due to a bad merge on my part https://github.com/supaglue-labs/supaglue/commit/5e78708ec4afd9275a6e96c6949291328e72ee68

Sorry guys! And thank you @lucasmarshall for catching it!

PR #1991, opened 2023-11-28
fix: don't allow customer deletions if there are any active connections

asdfryan:


tonyxiao (approved):

tonyxiao:
Was this the cause of the unable to delete issue?

asdfryan:
> Was this the cause of the unable to delete issue?

Yes. It turns out we do _attempt_ to clean up connections, but hit a snag if there are any existing syncs (for which we don't have the cascade). I think it's better to just not allow deleting customers with active connections.

PR #1995, opened 2023-11-28
fix: slightly better banner scroll behavior

tomkit:
- move banner to app area only to fix scroll behavior between navigator and app
Screenshot 2023-11-28 at 3 18 33 PM
Screenshot 2023-11-28 at 3 18 49 PM


lucasmarshall (commented):
lint check is failing

lucasmarshall (approved):

PR #2005, opened 2023-11-30
fix: revert "feat!: URL-encode cursors"

asdfryan:
Reason: for developers using http clients (like axios) this may result in double-url-encoding which will lead to 500s.

### DIFF
--- a/packages/core/lib/pagination.ts
+++ b/packages/core/lib/pagination.ts
@@ -48,7 +48,7 @@ export type Cursor = {
 };
 
 export const encodeCursor = (cursorParams: Cursor): string => {
-  return encodeURIComponent(Buffer.from(JSON.stringify(cursorParams), 'binary').toString('base64'));
+  return Buffer.from(JSON.stringify(cursorParams), 'binary').toString('base64');
 };
 
 export const decodeCursor = (encoded?: string): Cursor | undefined => {
--- a/packages/core/lib/pagination.ts
+++ b/packages/core/lib/pagination.ts
@@ -1,4 +1,5 @@
 import type { PaginationInternalParams, PaginationParams, ProviderName } from '@supaglue/types';
+import base64url from 'base64url';
 import { BadRequestError } from '../errors';
 
 export const DEFAULT_PAGE_SIZE = 10;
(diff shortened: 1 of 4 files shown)


packages/core/lib/pagination.ts
```diff
@@ -48,7 +48,7 @@ export type Cursor = {
 };
 
 export const encodeCursor = (cursorParams: Cursor): string => {
-  return encodeURIComponent(Buffer.from(JSON.stringify(cursorParams), 'binary').toString('base64'));
+  return Buffer.from(JSON.stringify(cursorParams), 'binary').toString('base64');
```
tonyxiao, line 51: So base64 encoding is always URL safe?

tonyxiao: From  chatGPT 
> Base64 encoding is generally considered URL safe. However, the standard Base64 encoding may contain characters that have special meaning in URLs (such as "+" and "/"), which could cause issues when used in a URL. To address this, a URL-safe variant of Base64 encoding, often called "URL encoding" or "base64url encoding," replaces these characters with URL-safe alternatives ("-_" instead of "+/"). Therefore, if you want to ensure compatibility with URLs, it's recommended to use the URL-safe variant of Base64 encoding.

if i remember base64 usually contains things like `=` that aren't the best in urls

asdfryan: It's not. 

However, if we return URL-encoded cursors, and they are using an HTTP Client that does URL-encoding for them (e.g. axios), the result is that it will become double-encoded and they will get a 500 back. This is what I believe is happening to one of our customers.

We had a discussion in the office about this. We should optimize for the code-path and not the curl/postman path.

tonyxiao: What about a base64 variant that is url safe?

asdfryan: We could use base58 encoding I suppose, but I also think a developer having to retry a request after encoding params is not a big deal. I often have to deal with this with date times etc.

Again, this would not affect most codepaths that already use a higher-level HTTP client like axios, which will URL encode things for them anyway.

tonyxiao: https://github.com/brianloveswords/base64url for example

lucasmarshall: Or we could upgrade to the Node 20 (the current LTS release) and `base64url` is a built-in supported encoding for `Buffer`s

tonyxiao: didn't know about base58, cool! Yea I think would be better to use either modified base64 or base58. But not blocking either way.

tonyxiao: and +1 for built in base64url encoding :) Is node 20 a big upgrade for us?

tonyxiao (approved):

packages/core/lib/pagination.ts
lucasmarshall: It would need testing, so we probably should just use a library for now and move to the built in after we upgrade.

PR #2013, opened 2023-12-01
feat!: unified error interface

lucasmarshall:


tonyxiao (approved):
Looks good. I wish we can have the codgen artifacts out of git cuz they make it super hard to focus on what actually happened! 20k lines of change is a ton of clutter! I looked through the changes in apps/ directory and packages/ and they both look good

PR #2024, opened 2023-12-04
feat: always upsert sequence state

asdfryan:
Tested locally


openapi/v2/engagement/paths/sequence_states.yaml
```diff
@@ -2,7 +2,7 @@ post:
   operationId: createSequenceState
   summary: Create sequence state
   description: |
-    In other words, adding a sequencestate to sequence.
+    In other words, adding a contact to sequence. If the contact is already in the sequence, it will return the id of the existing sequence state.
```
tomkit, line 5: ```Add a contact to a sequence. If the contact is already in the sequence, it will return the id of the existing sequence state.```

tomkit (approved):
nit: some grammarly fixes

PR #2025, opened 2023-12-04
feat(mgmt-ui): Add per-object overrides to mgmt UI

asdfryan:
Tested locally

https://www.loom.com/share/601ba5b4061544fdbeb2b8e92b4d291c


tomkit:
will play around with it

for the UI: can you nest/indent or add a different colored background to the parent element around all of the "object overrides" elements?

asdfryan:
image

lucasmarshall (approved):

PR #2033, opened 2023-12-06
feat: @supaglue/sdk first release

tonyxiao:


tonyxiao:
I'm thinking of changing `@supaglue/schemas` to simply `@supaglue/sdk`, it's not technically 100% correct naming given it contains a bunch of unrelated code but feels like the least friction overall and not much tech debt being introduced.

tonyxiao:
https://www.npmjs.com/package/@supaglue/sdk?activeTab=readme 

Need to add docs, but already works!

asdfryan (commented):
How much effort would it be to migrate our existing integration tests to use the SDK?

tonyxiao:
Not much. We already do in at least one integration test. Should be a straightforward refactor.

packages/sdk/package.json
```diff
@@ -1,9 +1,11 @@
 {
-  "name": "@supaglue/schemas",
+  "name": "@supaglue/sdk",
```
lucasmarshall, line 2: Have we tested this in a non-monorepo project? We should have to set `"exports"` at least so that we can specify where the `index.js` file is.

lucasmarshall (changes requested):

packages/sdk/package.json
tonyxiao: yea, I tested it via grabbing the published npm package and confirm it works.

PR #2036, opened 2023-12-06
feat!: surface default and override sync strategies

tomkit:
- surface frequency, period full, and strategy on sync page (respecting inheritance), associations to sync
- breaking change: snakecased `related_sync_states`
- removed entity names from syncs so there's more room

Screenshot 2023-12-06 at 5 00 03 PM

note: we should move to using classes for Supaglue's core concepts for better encapsulation (we have a lot of loose functions in different places)

## Test Plan

- tested locally

## Deployment instructions

### DIFF
--- a/apps/api/routes/mgmt/v2/sync/index.ts
+++ b/apps/api/routes/mgmt/v2/sync/index.ts
@@ -21,6 +21,7 @@ import type {
   TriggerSyncResponse,
 } from '@supaglue/schemas/v2/mgmt';
 import type { ObjectSyncDTO } from '@supaglue/types/sync';
+import { snakecaseKeys } from '@supaglue/utils/snakecase';
 import type { Request, Response } from 'express';
 import { Router } from 'express';
 
@@ -68,7 +69,7 @@ export default function init(app: Router) {
 
       const objectSyncs = results.filter((result) => result.type === 'object') as ObjectSyncDTO[];
 
-      const snakeCaseResults = results.map((result) => {
+      const snakecaseResults = results.map((result) => {
         const base = {
           id: result.id,
           connection_id: result.connectionId,
@@ -83,7 +84,9 @@ export default function init(app: Router) {
             type: 'object' as const,
             object_type: result.objectType,
             object: result.object,
-            related_sync_states: req.query?.customer_id ? generateRelatedObjectSyncStates(objectSyncs) : {},
+            related_sync_states: req.query?.customer_id
+              ? snakecaseKeys(generateRelatedObjectSyncStates(objectSyncs))
+              : {},
           };
         }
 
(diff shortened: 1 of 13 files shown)


apps/mgmt-ui/src/components/logs/SyncsTable.tsx
```diff
@@ -72,21 +70,75 @@ export default function SyncsTable(props: SyncsTableProps) {
...
+
+        if (activeSyncStrategy.fullSyncEveryNIncrementals !== undefined) {
+          return <span>every {activeSyncStrategy.fullSyncEveryNIncrementals} incremental</span>;
+        }
+
+        return <span>No</span>;
+      },
+    },
+    {
+      field: '_autoStartOnConnection',
```
asdfryan, line 120: This is not a property of the Sync and rather the SyncConfig so should be omitted.

What I mean is, it only affects future Syncs that have not been created and has no bearing on the individual Sync.

apps/mgmt-ui/src/components/logs/SyncsTable.tsx
```diff
@@ -72,21 +70,75 @@ export default function SyncsTable(props: SyncsTableProps) {
       filterOperators: equalOperatorOnly,
     },
     {
-      field: 'entityId',
-      headerName: 'EntityId',
-      valueGetter: (params) =>
-        params.row.entityId
-          ? entities.find((entity) => entity.id === params.row.entityId)?.name ?? params.row.entityId
-          : '',
+      field: '_frequency',
```
asdfryan, line 73: I thought you were also going to display associations to fetch?

apps/mgmt-ui/src/components/logs/SyncsTable.tsx
tomkit: Good point, I'll remove it

apps/mgmt-ui/src/components/logs/SyncsTable.tsx
```diff
@@ -72,21 +70,75 @@ export default function SyncsTable(props: SyncsTableProps) {
...
+      field: '_strategy',
+      headerName: 'Strategy',
+      width: 140,
+      sortable: false,
+      filterable: true,
+      filterOperators: equalOperatorOnly,
+      renderCell: (params) => {
+        const sync = params.row;
+        const { syncConfig } = params.row;
+        const activeSyncStrategy = getSyncStrategyConfig(syncConfig, sync);
```
asdfryan, line 96: We should convert this to the human readable version instead of the machine one. `full then incremental` -> `incremental` and `full only` -> `full`

apps/api/routes/mgmt/v2/sync/index.ts
```diff
@@ -76,6 +76,7 @@ export default function init(app: Router) {
           customer_id: result.customerId,
           sync_config_id: result.syncConfigId,
           paused: result.paused,
+          sync_config: result.syncConfig,
```
asdfryan, line 79: Doesn't this affect the openAPI spec?

I feel like instead of changing the API we should just fetch the sync configs in the frontend (there's <5 of them per application, probably), and then do the mapping there.

asdfryan (approved):

PR #2040, opened 2023-12-07
feat: implement hubspot webhook target endpoint + webhook setup

asdfryan:
- Create webhook target on Provider creation
- Create / update webhook subscriptions on SyncConfig update


packages/core/lib/hubspot_webhook.ts
```diff
@@ -0,0 +1,82 @@
+import { Client as HubspotClient } from '@hubspot/api-client';
+import type { SubscriptionCreateRequestEventTypeEnum } from '@hubspot/api-client/lib/codegen/webhooks';
+import { BadRequestError } from '../errors';
+
+const HUBSPOT_WEBHOOK_TARGET_URL =
+  process.env.HUBSPOT_WEBHOOK_TARGET_URL ?? `${process.env.SUPAGLUE_SERVER_URL}/internal/hubspot/_webhook`;
```
lucasmarshall, line 6: When would you use `HUBSPOT_WEBHOOK_TARGET_URL` instead of relying on the fallback?

packages/core/lib/hubspot_webhook.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const checkWebhookTargetExists = async (hubspotClient: HubspotClient, hubspotAppId: number): Promise<boolean> => {
+  try {
+    const res = await hubspotClient.webhooks.settingsApi.getAll(hubspotAppId);
+    if (res.targetUrl === HUBSPOT_WEBHOOK_TARGET_URL) {
+      return true;
+    }
+    throw new BadRequestError(
+      `Your Hubspot Developer App already has an existing Webhook target URL. Please delete it first.`
+    );
```
lucasmarshall, line 56: Maybe move this out of the `try` so we aren't needlessly catching and re-throwing it?

packages/core/lib/hubspot_webhook.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+const checkWebhookTargetExists = async (hubspotClient: HubspotClient, hubspotAppId: number): Promise<boolean> => {
+  try {
+    const res = await hubspotClient.webhooks.settingsApi.getAll(hubspotAppId);
+    if (res.targetUrl === HUBSPOT_WEBHOOK_TARGET_URL) {
+      return true;
+    }
+    throw new BadRequestError(
+      `Your Hubspot Developer App already has an existing Webhook target URL. Please delete it first.`
```
lucasmarshall, line 55:
```suggestion
      `Your Hubspot Developer App already has an existing Webhook target URL. Please delete it first or use a different app.`
```

lucasmarshall (changes requested):

packages/core/lib/hubspot_webhook.ts
asdfryan: When we use another service for queueing webhooks, like hookdeck.

packages/core/lib/hubspot_webhook.ts
```diff
@@ -0,0 +1,82 @@
...
+      return true;
+    }
+  } catch (e: any) {
+    if (e.code === 404) {
+      return false;
+    }
+    throw e;
+  }
+  throw new BadRequestError(
+    `Your Hubspot Developer App already has an existing Webhook target URL. Please delete it first.`
```
lucasmarshall, line 61:
```suggestion
    `Your Hubspot Developer App already has an existing Webhook target URL. Please delete it first or use a different Developer App.`
```

lucasmarshall (approved):
Small nit, otherwise looks good.

PR #2051, opened 2023-12-12
feat: expand associations for single GET (salesforce + hubspot only)

asdfryan:
This is part 1 of expanding associations.

This also includes a large refactor to move `includeRawData` to the service / provider layers.

This also fixes a bug where we were performing schema mapping twice for managed data. (The synced data is already mapped, so no need to map it again)

TODO:
- implement expand for uncached LIST
- implement expand for cached LIST
- apply schemas on expanded objects


apps/api/routes/crm/v2/account.ts
```diff
@@ -42,46 +40,35 @@ export default function init(app: Router): void {
...
+      // TODO: Implement expand for uncached reads
+      if (req.query?.expand?.length) {
+        throw new BadRequestError('Expand is not yet supported for uncached reads');
       }
-      return res.status(200).send({
-        pagination,
-        records: records.map((record) => ({
-          ...record,
-          raw_data:
-            includeRawData && fieldMappingConfig ? toMappedProperties(record.raw_data, fieldMappingConfig) : undefined,
```
asdfryan, line 78: We are reading synced data, which is already mapped. No need to map it again.

apps/api/routes/crm/v2/account.ts
```diff
@@ -42,46 +40,35 @@ export default function init(app: Router): void {
       if (req.query?.read_from_cache?.toString() !== 'true') {
         const { pagination, records } = await crmCommonObjectService.list('account', req.customerConnection, {
           modifiedAfter: req.query?.modified_after,
+          expand: req.query?.expand?.split(','),
```
lucasmarshall, line 43: We should be able to not have to do this if we set up the query string parser correctly:

```
import qs from `qs`;
app.set('query parser', function (str) {
  return qs.parse(str, { comma: true });
});
```

lucasmarshall (approved):

PR #2053, opened 2023-12-12
chore: remove unnessessary heartbeating readable code

lucasmarshall:
No longer required as we already heartbeat on a 10 second interval while syncs are running.


asdfryan (commented):
As discussed, we should emit heartbeats regularly on the read portion of the workflow as well.

lucasmarshall:
> As discussed, we should emit heartbeats regularly on the read portion of the workflow as well.

This actually does this already. `writeObjects` does the reads and writes, and we set up the heartbeating before we call that.

asdfryan (approved):

PR #2056, opened 2023-12-12
fix: return a 401 when tokens can't be refreshed in API

lucasmarshall:


packages/core/errors.ts
```diff
@@ -214,6 +214,6 @@ export class SGTerminalTooManyRequestsError extends SGSyncWorkerError {
 // Throw this when you want run_object_sync to pause after the current execution
 export class SGConnectionNoLongerAuthenticatedError extends SGSyncWorkerError {
```
lucasmarshall, line 215: This is also being thrown in API, so we need to have the correct http status code for it

asdfryan (approved):

